Issue #2, opened 2024-06-22
`accounts: Vec<_>`


cavemanloverboy:
Why is this a vector?

https://github.com/buffalojoec/mollusk/blob/e861087f3d4f61fcbb942fb056d701f068c60822/harness/src/lib.rs#L158-L162

Should make this a slice so that devs don't have to clone (and so you don't have to clone internally on each loop iteration).

cavemanloverboy:
I can submit PR if you agree, just say the word

buffalojoec:
Yep, definitely agree! I made this a slice originally, but since `TransactionContext` takes a vector, I didn't manage to see a good way around cloning within the harness, so I left it up to the developer to decide if they wanted to clone when providing the accounts.

If you can make it work without cloning, I'm game!

cavemanloverboy:
Even if you end up doing copies internally, It's one less copy (devs have to clone the vector due to your interface and then it's internally copied onto the `transaction_accounts` buffer again). And, I think it's a nicer API.

buffalojoec:
> Even if you end up doing copies internally, It's one less copy (devs have to clone the vector due to your interface and then it's internally copied onto the `transaction_accounts` buffer again). And, I think it's a nicer API.

Works for me! Thanks!

cavemanloverboy:
https://github.com/buffalojoec/mollusk/pull/6

PR #6, opened 2024-06-22
accounts: use slice instead of vec

cavemanloverboy:
addresses #2

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -158,7 +158,7 @@ impl Mollusk {
     pub fn process_instruction(
         &self,
         instruction: &Instruction,
-        accounts: Vec<(Pubkey, AccountSharedData)>,
+        accounts: &[(Pubkey, AccountSharedData)],
     ) -> InstructionResult {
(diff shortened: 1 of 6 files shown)


harness/src/lib.rs
```diff
@@ -177,8 +177,9 @@ impl Mollusk {
             .collect::<Vec<_>>();
 
         let transaction_accounts = [(self.program_id, self.program_account.clone())]
-            .into_iter()
+            .iter()
             .chain(accounts)
+            .cloned()
```
buffalojoec, line 182: Yeah, this was exactly what I had. Unfortunate, but I agree the API is better served with a reference.

buffalojoec (commented):
Lgtm, thanks! Just need that one touch-up to pass clippy (I think I broke the other one).

Since you've added the lifetime specifier to `Bench`, in a follow-up PR I'm going to add some more references there.

cavemanloverboy:
fixed clippy

buffalojoec (approved):
Thanks!

PR #19, opened 2024-06-26
Add BPF Loader 2 to cache

buffalojoec:
Add the BPF Loader 2 to the default program cache. This allows Mollusk
to support CPI to BPF Loader 2 programs.

cc @OliverNChalk

### DIFF
--- a/harness/src/program.rs
+++ b/harness/src/program.rs
@@ -40,7 +40,7 @@ impl ProgramCache {
         &self.cache
     }
 
-    /// Add a program to a program cache.
+    /// Add a program to the cache.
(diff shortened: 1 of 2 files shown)


buffalojoec:
Just a general note for anyone following.

Rather than add all of the builtins to the cache, like the runtime does, I'd rather not add unnecessary dependencies to Mollusk - like the lookup table program, or stake/vote. Instead, if developers need those builtins, they can add those crates to their development dependencies and add them themselves through Mollusk's `program` API.

All builtins besides system and BPF Loaders are thus opt-in only.

PR #43, opened 2024-11-27
create EpochSchedule without warmup

2501babe:
`EpochSchedule::default()` sets a warmup period, which means you need to do things to figure out what the first normal slot and epoch are. by using `EpochSchedule::without_warmup()` instead, the first normal slot and epoch are both 0. this makes typical usage easier. tests that want to see a warmup can override `EpochSchedule` explicitly


buffalojoec (approved):
tyvm!

PR #44, opened 2024-11-27
reconstruct slot hashes for long warps

2501babe:
as-configured an epoch is 432k slots, so jumping more than a couple epochs takes several seconds of runtime because `warp_to_slot` must binary search, insert, and truncate `SlotHashes` potentially millions of times. since `SlotHashes` can only contain 512 entries, we can reconstruct it from scratch ~instantly


buffalojoec (approved):
Nice, I love it. Solid change.

PR #45, opened 2024-11-28
init stake history with epoch

2501babe:
stake history is segmented by epoch rather than slot. this works because by default both are 0, but isnt correct

### DIFF
--- a/harness/src/sysvar.rs
+++ b/harness/src/sysvar.rs
@@ -44,7 +44,7 @@ impl Default for Sysvars {
         };
 
         let mut stake_history = StakeHistory::default();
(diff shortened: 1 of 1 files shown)


harness/src/sysvar.rs
```diff
@@ -44,7 +44,7 @@ impl Default for Sysvars {
         };
 
         let mut stake_history = StakeHistory::default();
-        stake_history.add(clock.slot, StakeHistoryEntry::default());
+        stake_history.add(clock.epoch, StakeHistoryEntry::default());
```
buffalojoec, line 47: Oops, lol.

buffalojoec (approved):

PR #60, opened 2024-12-04
Support Solana minor versions

kox:
Based on the current cargo dependencies, it's a bit confusing which minor versions aresupported per each solana package. 

This PR intends to give support to all 2.1 minor versions and avoid conflicts when using latest version. 

Feel free to close the PR if you don't find it convenient or due security or compatibility. 

Cheers


kox:
closing PR due possible incompatibilities

PR #61, opened 2024-12-06
override sysvars from accounts list

2501babe:
this is a convenience to make porting stake_instruction.rs tests (and other similar `InvokeContext`-based tests) much simpler. sysvars from the accounts list are used to override the sysvars on `Mollusk`, without mutating it. this is to accord with the way existing tests are structured, so we dont need to add any setup logic that doesnt already exist. in general they create sysvar states and pass them into `InvokeContext` via `process_instruction` directly 

i considered letting it mutate `Sysvars` but this seems cleaner, especially since we will be able to remove it someday when the bpf programs use sysvar syscalls and dont need to accept sysvar accounts at all

### DIFF
--- a/harness/src/fuzz/firedancer.rs
+++ b/harness/src/fuzz/firedancer.rs
@@ -133,10 +133,13 @@ fn parse_fixture_context(
         .map(|(key, acct, _)| (*key, acct.clone()))
         .collect::<Vec<_>>();
 
+    let mut sysvars = Sysvars::default();
+    sysvars.fill_from_accounts(&accounts);
+
     let mollusk = Mollusk {
         compute_budget,
         feature_set: epoch_context.feature_set.clone(),
-        sysvars: Sysvars::fill_from_accounts(&accounts),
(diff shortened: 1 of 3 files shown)


buffalojoec (commented):
I think this manages to accomplish what you're after here: https://github.com/buffalojoec/mollusk/pull/64

Let me know if there's more we need!

2501babe:
at a glance this looks right! ill let you know after rebases if i have any issues

fwiw this doesnt support all firedancer fixtures, because a number of them (iirc 1% of stake fixtures) have unparseable sysvar data. i dont think these fixtures really make any sense, since such cases are impossible (at least, would be indicative of whole-network corruption) and we shouldnt change `Sysvars` to hold `Option<T>`s to support them, but just making you aware

when i dealt with these i just wrapped the builder in `panic::catch_unwind` and skipped anything that failed to parse. if/when we support their fixtures in our ci we should probably just petition for these to be removed rather than do anything in mollusk to support them

PR #62, opened 2024-12-06
add native stake program

2501babe:
this makes it easy to write tests that use `test_case` to run against both native and bpf stake to confirm the results are the same

i wasnt entirely use on naming `stake_program` vs `solana_stake_program`, i think the latter is what we prefer now, and we might want to rename `system_program` to `solana_system_program`


buffalojoec (commented):
Do you mind putting this behind a feature flag? How about we just use `all-builtins`, and I can stick other builtins in here too later, like vote.

2501babe:
i didnt add `system_program` to the feature since the bpf test program depends on it, i think this is fine until we are working on a bpf system program

buffalojoec (approved):
Yep, works for me, thanks!

PR #63, opened 2024-12-06
add support for return data

2501babe:
using this for minimum delegation tests with normal mollusk, i didnt modify any of the fuzz types since im not sure what support is needed for those

### DIFF
--- a/harness/src/fuzz/firedancer.rs
+++ b/harness/src/fuzz/firedancer.rs
@@ -232,6 +232,7 @@ fn parse_fixture_effects(
             .compute_budget
             .compute_unit_limit
             .saturating_sub(effects.compute_units_available),
+        return_data: None, // TODO: Omitted for now.
         resulting_accounts,
     }
 }
--- a/harness/src/fuzz/firedancer.rs
+++ b/harness/src/fuzz/firedancer.rs
@@ -169,6 +169,8 @@ fn build_fixture_effects(context: &FuzzContext, result: &InstructionResult) -> F
         }
     };
 
+    let return_data = result.return_data.clone();
+
     let modified_accounts = context
         .accounts
         .iter()
@@ -191,7 +193,7 @@ fn build_fixture_effects(context: &FuzzContext, result: &InstructionResult) -> F
         compute_units_available: context
             .compute_units_available
             .saturating_sub(result.compute_units_consumed),
-        return_data: Vec::new(), // TODO: Mollusk doesn't capture return data.
+        return_data,
     }
 }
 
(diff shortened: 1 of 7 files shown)


harness/src/fuzz/mollusk.rs
```diff
@@ -94,6 +94,7 @@ impl From<&FuzzEffects> for InstructionResult {
             execution_time,
             program_result,
             raw_result,
+            return_data: None, // TODO: Omitted for now.
```
buffalojoec: Can you add this field to the non-Firedancer protobuf and fixture lib? It should be pretty simple, let's just do `Vec<u8>` like FD does.

Now that the Mollusk result type tracks return data, the Mollusk-based fixture effects should also.

harness/src/fuzz/firedancer.rs
```diff
@@ -235,6 +235,7 @@ fn parse_fixture_effects(
             .compute_budget
             .compute_unit_limit
             .saturating_sub(effects.compute_units_available),
+        return_data: None, // TODO: Omitted for now.
```
buffalojoec: Have you seen fixtures in the Firedancer test-vectors that test the return data for Stake? If so, we might want to add this to the Firedancer support here.

I can get a test going that actually runs against real test vectors to make sure this thing actually works, and you could rebase onto it if you want.

buffalojoec (commented):
Change looks good, but I think we need to extend the support to the fuzzing features.

harness/src/fuzz/firedancer.rs
buffalojoec: Here's the PR: https://github.com/buffalojoec/mollusk/pull/65

2501babe: yea a lot of the fixtures contain expected return data for `GetMinimumDelegation`, i just havent looked through the fuzz and fd-fuzz code yet to see what support is needed since my usecase only depends on harness. will do that tho

buffalojoec: You shouldn't even have to mess with the fuzz-related crates, the change(s) should be local to the main harness's `fuzz` dir.

But lmk if you want me to land my stuff first, so we have the coverage.

2501babe:
ok it took me a couple tries but i think i got it now haha

harness/src/fuzz/firedancer.rs
```diff
@@ -191,7 +193,7 @@ fn build_fixture_effects(context: &FuzzContext, result: &InstructionResult) -> F
         compute_units_available: context
             .compute_units_available
             .saturating_sub(result.compute_units_consumed),
-        return_data: Vec::new(), // TODO: Mollusk doesn't capture return data.
```
buffalojoec, line 194: Niceeee

buffalojoec (approved):
Looks perfect. Thank you!!

PR #70, opened 2024-12-20
harness: Add data slice check

febo:
### Problem

Currently there is not a quick way to check the value of a field against a known value after an instruction is executed.

### Solution

Add a `data_slice` check to `AccountCheck`. A test against a data slice is done by specified the start offset and the slice data – e.g.,:
```rust
Check::account(&mint_key)
    .data_slice(50, mint_authority_key.as_ref())
    .build()
```


buffalojoec (approved):
Lgtm, thanks!

buffalojoec:
I rebased this onto #71 so it would pass CI. Merging!

PR #74, opened 2025-01-09
deps: Unpin Solana dependencies

joncinque:
#### Problem

It's very useful to run downstream tests with newer versions of the Solana crates to catch unwanted breakage. Mollusk pins some of its Solana crate dependencies to 2.1.0.

The token repo uses mollusk for its testing framework, but because of the pinning in mollusk, it's impossible to run downstream tests on token with newer versions of the monorepo.

#### Summary of changes

This might be totally off-base, and I might be missing something important, but it'll be way more flexible and easier if mollusk unpins its dependencies.


buffalojoec (commented):
The only thing we have to be careful of here is the fact that APIs within `solana-program-runtime` do not respect semver, and therefore can change and break Mollusk.

We can ship this for the main release line, but I think I'll also cut a `solana-2.1` release line where versions are aligned with Solana/Agave versions, in case any API changes break.

For example, I know a few fields on `ComputeBudget` changed between some 2.1.x and 2.1.y versions.

buffalojoec:
> Gotcha, it's your library, so I'll let you decide the best course of action. I'm surprised there were breaking changes in a patch release 😞

Yeah it's too bad, but let's ship it 🚢 and see what happens! We might be far enough along at this point.

PR #75, opened 2025-01-12
fuzz: `program_result` as `u64`

febo:
### Problem

Currently, the `program_result` in the struct`Effects` is a `u32` value. But the built-in `InstructionError` values occupy the upper 32 bits – casting the resulting `u64` to `u32` effectively drops their value. The issue is apparent when the instruction fails with a `ProgramError::Custom(0)` and fixtures are generated. The result of the cast in this case be `0`, which is the same value as the `ProgramResult::Success`:
```rust
let program_result = match &program_result {
        ProgramResult::Success => 0,
        ProgramResult::Failure(e) => u64::from(e.clone()) as u32,
        ProgramResult::UnknownError(_) => u32::MAX, //TODO
};
```

### Solution

Update the value of `program_result` to be a `u64`.

### DIFF
--- a/fuzz/fixture/proto/invoke.proto
+++ b/fuzz/fixture/proto/invoke.proto
@@ -62,7 +62,7 @@ message InstrEffects {
     uint64 execution_time = 2;
 
     // Program return code. Zero is success, errors are non-zero.
-    uint32 program_result = 3;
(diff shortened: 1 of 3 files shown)


buffalojoec (approved):
Thanks!!

PR #77, opened 2025-01-13
fuzz: refactor fixture processing harnesses to use `&mut self`

buffalojoec:
When working with fixtures, it's sometimes necessary to configure Mollusk's program cache before running some set of fixtures.

In the current implementation, functions such as `process_and_validate_fixture` are associated functions, so they vend their own new `Mollusk` instance internally and use that for processing the fixture, giving the developer no opportunity to configure the program cache (or any other configurations).

This change makes these functions _methods_ using `&mut self`, which causes the loaded fixture to update the current `Mollusk` instance's configurations based on the fixture.

A few notes:
* For any developers who do _not_ wish to alter their `Mollusk` instance based on loaded fixtures, the `load_fixture` function will no longer vend a `Mollusk` instance, but the components themselves, allowing developers to do with the components what they wish.
* The `run_checks` method on the `InstructionResult` type has been made public, in case comparing all checks in `process_and_validate_fixture` (or any other Mollusk method for that matter) is too strict. Now, one could run just a subset of checks on the returned result from, say, `process_fixture`.

cc @febo

### DIFF
--- a/harness/src/fuzz/check.rs
+++ b/harness/src/fuzz/check.rs
@@ -1,0 +1,173 @@
+//! Checks to run against a fixture when validating.
+
+use {
+    crate::result::{Check, InstructionResult},
+    solana_sdk::{
+        account::{Account, ReadableAccount},
+        pubkey::Pubkey,
+    },
+};
+
+/// Checks to run against a fixture when validating.
+///
+/// Similar to Mollusk's `result::Check`, this allows a developer to dictate
+/// the type of checks to run on the fixture's effects.
+///
+/// Keep in mind that validation of fixtures works slightly differently than
+/// typical Mollusk unit tests. In a Mollusk test, you can provide the value to
+/// compare a portion of the result against (ie. compute units). However, when
+/// comparing the result of a Mollusk invocation against a fixture, the value
+/// from the fixture itself is used.
+///
+/// For that reason, these are unary variants, and do not offer the developer a
+/// way to provide values to check against.
+pub enum FixtureCheck {
+    /// Validate compute units consumed.
+    ComputeUnits,
+    /// Validate the program result.
+    ProgramResult,
+    /// Validate the return data.
+    ReturnData,
+    /// Validate all resulting accounts.
+    AllResultingAccounts {
+        /// Whether or not to validate each account's data.
+        data: bool,
+        /// Whether or not to validate each account's lamports.
+        lamports: bool,
+        /// Whether or not to validate each account's owner.
+        owner: bool,
+        /// Whether or not to validate each account's space.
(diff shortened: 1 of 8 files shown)


harness/src/fuzz/check.rs
```diff
@@ -0,0 +1,173 @@
...
+/// way to provide values to check against.
+pub enum FixtureCheck {
+    /// Validate compute units consumed.
+    ComputeUnits,
+    /// Validate the program result.
+    ProgramResult,
+    /// Validate the return data.
+    ReturnData,
+    /// Validate all resulting accounts.
+    AllResultingAccounts {
```
febo, line 32: Would it make sense to have a variant without any data representing the case where you want to check everything from the resulting accounts? It would be a simpler way to do that (no need to add the `bool` values) and it might be the common case.

Not so sure about the naming that we could have, but maybe something along the lines:
```
...
/// Validates all resulting accounts.
ResultingAccounts,
/// Validates specific fields of all resulting accounts.
ResultingAccountsFields {
        /// Whether or not to validate each account's data.
        data: bool,
        /// Whether or not to validate each account's lamports.
        lamports: bool,
        /// Whether or not to validate each account's owner.
        owner: bool,
        /// Whether or not to validate each account's space.
        space: bool,
},
/// Validates specific fields of a subset resulting accounts.
ResultingAccountsSubset {
        /// The addresses on which to apply the validation.
        addresses: Vec<Pubkey>,
        /// Whether or not to validate each account's data.
        data: bool,
        /// Whether or not to validate each account's lamports.
        lamports: bool,
        /// Whether or not to validate each account's owner.
        owner: bool,
        /// Whether or not to validate each account's space.
        space: bool,
},
/// Validates specific fields of resulting accounts _except_ the provided addresses.
ResultingAccountsExcept {
        /// The addresses on which to _not_ apply the validation.
        ignore_addresses: Vec<Pubkey>,
        /// On non-ignored accounts, whether or not to validate each account's
        /// data.
        data: bool,
        /// On non-ignored accounts, whether or not to validate each account's
        /// lamports.
        lamports: bool,
        /// On non-ignored accounts, whether or not to validate each account's
        /// owner.
        owner: bool,
        /// On non-ignored accounts, whether or not to validate each account's
        /// space.
        space: bool,
    },
}

febo (commented):
Looks great! Left a small suggestion.

febo:
Run a test with the `FixtureCheck` and they work fine.

harness/src/fuzz/check.rs
buffalojoec: Added helpers for this in my most recent commit. Nice suggestion!

PR #73, opened 2025-01-03
feat: support fixture generation in instruction chains

buffalojoec:
This change adjusts the internals of the instruction chain API to support generation of fuzzing fixtures.

Additionally, since checks will play a key role in fixtures down the road, this change also adjusts the API for instruction chains, allowing developers to declare `ChainChecks` on their instruction chain validation, which can specify a certain index for which instruction to apply a set of checks.

Example from `harness/tests/instruction_chain.rs`:

```rust
mollusk.process_and_validate_instruction_chain(
    &[
        system_instruction::transfer(&alice, &bob, alice_to_bob),
        system_instruction::transfer(&bob, &carol, bob_to_carol),
        system_instruction::transfer(&bob, &dave, bob_to_dave),
    ],
    &[
        (alice, system_account_with_lamports(starting_lamports)),
        (bob, system_account_with_lamports(starting_lamports)),
        (carol, system_account_with_lamports(starting_lamports)),
        (dave, system_account_with_lamports(starting_lamports)),
    ],
    &[
        ChainChecks::At(
            0, // 0: Alice to Bob
            &[
                Check::success(),
                Check::account(&alice)
                    .lamports(starting_lamports - alice_to_bob) // Alice pays
                    .build(),
                Check::account(&bob)
                    .lamports(starting_lamports + alice_to_bob) // Bob receives
                    .build(),
                Check::account(&carol)
                    .lamports(starting_lamports) // Unchanged
                    .build(),
                Check::account(&dave)
                    .lamports(starting_lamports) // Unchanged
                    .build(),
            ],
        ),
        ChainChecks::At(
            1, // 1: Bob to Carol
            &[
                Check::success(),
                Check::account(&alice)
                    .lamports(starting_lamports - alice_to_bob) // Unchanged
                    .build(),
                Check::account(&bob)
                    .lamports(starting_lamports + alice_to_bob - bob_to_carol) // Bob pays
                    .build(),
                Check::account(&carol)
                    .lamports(starting_lamports + bob_to_carol) // Carol receives
                    .build(),
                Check::account(&dave)
                    .lamports(starting_lamports) // Unchanged
                    .build(),
            ],
        ),
        ChainChecks::At(
            2, // 2: Bob to Dave
            &[
                Check::success(),
                Check::account(&alice)
                    .lamports(starting_lamports - alice_to_bob) // Unchanged
                    .build(),
                Check::account(&bob)
[...]

PR #79, opened 2025-01-13
support fuzz fixtures in ix chains

buffalojoec:
Reimplementation of #73 from [this suggestion](https://github.com/buffalojoec/mollusk/pull/73#pullrequestreview-2546863518) from @febo.

---

This change adjusts the internals of the instruction chain API to support generation of fuzzing fixtures.

Additionally, since checks will play a key role in fixtures down the road, this change also adjusts the API for instruction chains, allowing developers to declare a slice of `Check` for each instruction in the chain separately.


febo:
Looks great!

PR #82, opened 2025-01-20
Added custom logger

deanmlittle:
demuxing logs when using mollusk in a multi-threaded environment is really annoying. by enabling us to pass in a custom logger, we are able to handle log capture.


buffalojoec (approved):
Thanks!

PR #86, opened 2025-02-02
Fix crate vulnerabilities

rex4539:
https://rustsec.org/advisories/RUSTSEC-2025-0004
https://rustsec.org/advisories/RUSTSEC-2024-0344

Also updated yanked crate `futures-util`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -828,9 +828,9 @@ dependencies = [
 
 [[package]]
 name = "curve25519-dalek"
-version = "3.2.0"
+version = "3.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -4394,9 +4394,9 @@ dependencies = [
 
 [[package]]
 name = "zeroize"
-version = "1.8.1"
+version = "1.3.0"
```
buffalojoec, line 4397: This one looks like a downgrade. Is this intentional?

rex4539: Yes, it happens automatically on `cargo update curve25519-dalek@3.2.0`.

buffalojoec (approved):
Thanks!

PR #88, opened 2025-02-15
upgrade solana crates to 2.2

kevinheavey:
This is a blocker for the PR to remove solana-program from spl-token https://github.com/solana-program/token/pull/26


kevinheavey:
I've opened a PR in solana-sdk to fix the size_of issue in CI: https://github.com/anza-xyz/solana-sdk/pull/40

Alternatively we could use a more recent Rust version in the tests, as size_of entered the Rust prelude in 1.80 (hence why there was no such error in Agave CI)

kevinheavey:
I would guess this other failure is from removing references to feature IDs that no longer exist in code:
```
 thread 'test_dump_mollusk' panicked at harness/tests/dump_fixture.rs:56:5:
assertion `left == right` failed
  left: 225
 right: 226
```

harness/src/lib.rs
```diff
@@ -429,17 +429,6 @@ impl Default for Mollusk {
...
              solana_runtime::system_instruction_processor=trace",
         );
-        #[cfg(feature = "fuzz")]
-        let feature_set = {
-            // Omit "test features" (they have the same u64 ID).
-            let mut fs = FeatureSet::all_enabled();
-            fs.active
-                .remove(&solana_sdk::feature_set::disable_sbpf_v1_execution::id());
-            fs.active
-                .remove(&solana_sdk::feature_set::reenable_sbpf_v1_execution::id());
```
buffalojoec: These have to be in here, but they've been renamed to `sbpf_v0` instead.

https://github.com/anza-xyz/solana-sdk/blob/eee5cab08f441f5ed92411794299e6f416b4dfe6/feature-set/src/lib.rs#L871-L877

fuzz/fixture/src/feature_set.rs
```diff
@@ -6,10 +6,7 @@ use {
 };
 
 // Omit "test features" (they have the same u64 ID).
-pub static OMITTED_FEATURES: &[Pubkey] = &[
-    solana_sdk::feature_set::disable_sbpf_v1_execution::id(),
-    solana_sdk::feature_set::reenable_sbpf_v1_execution::id(),
-];
```
buffalojoec: Same here!

buffalojoec (commented):
Thanks! After this, am I able to roll a PR to get rid of `solana-sdk` in this library? Looks to me like we're good to go.

buffalojoec:
> I would guess this other failure is from removing references to feature IDs that no longer exist in code:
> 
> ```
>  thread 'test_dump_mollusk' panicked at harness/tests/dump_fixture.rs:56:5:
> assertion `left == right` failed
>   left: 225
>  right: 226
> ```

Yeah, it's because they both start with `TestFeature1111...`. For lack of a better solution I've been just omitting them from the fuzzing stuff until we really need them. See [my comment here](https://github.com/anza-xyz/mollusk/pull/88#discussion_r1958985622).

Cargo.toml
```diff
@@ -46,16 +46,16 @@ serde = "1.0.203"
...
-solana-timings = "2.1.0"
+solana-bpf-loader-program = "2.2.0"
+solana-compute-budget = "2.2.0"
+solana-log-collector = "2.2.0"
+solana-logger = "2.2.1"
+solana-program = "2.2.1"
+solana-program-runtime = "2.2.0"
+solana-stake-program = "2.2.0"
+solana-system-program = "2.2.0"
+solana-sdk = "2.2.1"
```
buffalojoec: These specifically must be `^2.1.1`?

PR #94, opened 2025-03-14
cli: improve verbose output

buffalojoec:
Improve the verbose output for the Mollusk CLI.

Additionally - since I accidentally committed it all together - make both CLI commands share the final output log, regardless of verbosity.

### DIFF
--- a/cli/src/runner.rs
+++ b/cli/src/runner.rs
@@ -214,6 +214,10 @@ impl Runner {
         println!();
         println!("[DONE][TEST RESULT]: {} failures", failures);
 
+        if failures > 0 {
+            std::process::exit(1);
+        }
+
         Ok(())
     }
 }


harness/src/result.rs
```diff
@@ -18,7 +18,7 @@ macro_rules! compare {
                 panic!("{}", msg);
             } else {
                 if $c.verbose {
-                    eprintln!("{}", msg);
+                    println!("{}", msg);
```
buffalojoec, line 21: This change is necessary since `eprintln!` goes to `stderr`, and if you're running a job with lots of `stdout` logs and one of these is thrown, it appears non-deterministically in the developer's console output. Therefore, even if it is technically an error case, it's better DevEx to land it in `stdout`.

febo (approved):
Looks good!

PR #98, opened 2025-03-19
cli: reorder ground and target parameters

buffalojoec:
This was driving me nuts, as it was very confusing to follow. Now, both commands treat the provided binary as the "target", and `run-test` has the additive "ground truth".


febo (approved):
Looks good!

PR #99, opened 2025-03-19
cli: support generating cu reports

buffalojoec:
Adds support for for generating a CU consumption report from the CLI, therefore from processing fixtures. Produces the same markdown file as a [CU bench test](https://github.com/solana-program/config/blob/5b7e1d775bdbcbba6e3df82afde923899793a0f3/program/benches/compute_units.md), where each bench run is the fixture name.


febo (approved):
Perfect! 👌

PR #101, opened 2025-04-01
Update agave dependencies

febo:
### Problem

Programs from the agave repository are now using their "own" precompiles and feature-set crates. This causes problems for mollusk since while `solana-precompiles` and `solana-feature-set` have similar types, they are considered different.

### Solution

Update mollusk to depend on `agave-precompiles` and `agave-feature-set` instead.


fuzz/fixture-fd/src/feature_set.rs
```diff
@@ -1,20 +1,20 @@
...
-    solana_feature_set::reenable_sbpf_v0_execution::id(),
+    agave_feature_set::disable_sbpf_v0_execution::id(),
+    agave_feature_set::reenable_sbpf_v0_execution::id(),
 ];
 
 impl From<ProtoFeatureSet> for FeatureSet {
     fn from(value: ProtoFeatureSet) -> Self {
         let mut feature_set = Self::default();
-        let mut inactive = std::mem::take(&mut feature_set.inactive);
+        let mut inactive = feature_set.inactive().clone();
```
febo, line 17: I am not sure whether this needs the "take" or not – cloning should have the same effect since the `feature_set` has just being created.

buffalojoec (approved):
I'm curious how this behaves now that SDK is separate from Agave. It probably won't be much smoother until they're on different majors, but let's YOLO this and see what happens. 🫠

Nagaprasadvr:
this is great , thank you ser

Issue #104, opened 2025-04-16
Feature Request: add cpi breakdown to bench results


redmannequin:
It would be nice to have a breakdown of how many CU are incurred by the "root" program and it's CPI(s) calls in the result.

example: 
| Name | CUs | Delta | my_program | cpi(s) |
|------|------|-------|-------|-------|
| create_config | 8388 | -52 | 3657 | 4731 |
| update_config | 528 | -- | 528 | -- |
| update_claim | 863 | - new - | 863 | -- |

buffalojoec:
Unfortunately, it's not possible for Mollusk to capture CPI-only compute units without modifying the `InvokeContext` API that comes from Agave. `InvokeContext::process_instruction` accepts a `&mut u64` for compute units consumed, and doesn't give you a way to discern where the consumption comes from.

We try to avoid making changes to Agave that aren't necessary for validator tasks. Maybe you can share more about what you're trying to measure/do and we can see about an alternate approach?

redmannequin:
Wouldn't it be possible by parsing the logs? Is there anyway to access the logs of an execution? 

I'm currently building/playing around with a dummy [framework] built on top of pinocchio, and I'm trying to benchmark my changes to see their impact and catch any regressions.

[framework]: https://github.com/redmannequin/sol-ez/tree/main

redmannequin:
I just thought of another neat use-case -- give an options of result format, add json to to the mix; this way you can easily write a CI step to auto bench and flag regressions. (this is just an idea, probably need's its own issue just dumping the idea here)

redmannequin:
After thinking about it, this doesn’t feel like the right place to be parsing logs. Might be better to just surface the logs alongside the execution result — gives more flexibility and keeps the crate lightweight. What are your thoughts? 

Do you happen to know if there are any structured Solana log parsers out there?

buffalojoec:
To me, parsing logs seems like it's rife for bugs. A program could just emit logs that spoof the bencher. I don't know why you would, but you could.

If you want to parse logs on your end, you can! You just need to query the log collector's messages and apply your parsing.

> Do you happen to know if there are any structured Solana log parsers out there?

Nope, not that I know of. But if you create one, feel free to share!

redmannequin:
Perfect, I've update the PR with some changes, adds a new function that doesn't write the file and returns the bench results as well as it adds the logs of a bench execution to the result type.

ananas-block:
[...]

PR #105, opened 2025-04-16
Allow In-Memory Benchmarks by Exposing Results and Skipping File Write

redmannequin:
## Problem 
Currenly the `MolluskComputeUnitBencher` doesn't allow run the bench without writing the file and returning the results. 

## Solution
- add `execute_without_write` which only runs the bench and returns the bench results
- make the `MolluskComputeUnitBenchResult` public as well as it fields 

## Notes
- #104

### DIFF
--- a/bencher/src/lib.rs
+++ b/bencher/src/lib.rs
@@ -65,7 +65,7 @@ use {
     solana_account::Account,
     solana_instruction::Instruction,
     solana_pubkey::Pubkey,
-    std::{path::PathBuf, process::Command, str::FromStr},
+    std::{path::PathBuf, process::Command},
 };
 
 /// A bench is a tuple of a name, an instruction, and a list of accounts.
@@ -114,9 +114,15 @@ impl<'a> MolluskComputeUnitBencher<'a> {
 
     /// Execute the benches.
     pub fn execute(&mut self) {
+        let out_dir = self.out_dir.clone();
         let table_header = Utc::now().to_string();
         let solana_version = get_solana_version();
-        let bench_results = std::mem::take(&mut self.benches)
+        let bench_results = self.execute_without_write();
+        write_results(&out_dir, &table_header, &solana_version, bench_results);
+    }
+
+    pub fn execute_without_write(&mut self) -> Vec<MolluskComputeUnitBenchResult> {
+        std::mem::take(&mut self.benches)
             .into_iter()
             .map(|(name, instruction, accounts)| {
                 let logger = self.mollusk.logger.clone();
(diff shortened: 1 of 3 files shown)


buffalojoec (commented):
Sorry for the delay on reviewing here!

I like the approach of using a new method to avoid the breaking change on `execute`, however, the additional field to `MolluskComputeUnitBenchResult` is a breaking change according to Rust's type safety.

You make a good point that right now, developers can't opt-out of the file write. The default behavior is to write the file to a `benches` folder in the root, which is honestly not great. So I think one solid change is to offer that ability to opt-out of file writes. Since the fields on the `MolluskComputeUnitBencher` struct are private, we have some flexibility here. Maybe we can just add a `skip_markdown()` method to the builder?

As for returning bench results, I wonder if you are better off just writing benchmarks with `Mollusk` directly. There's not really anything you're doing in this PR that necessitates the `MolluskComputeUnitBencher` over `Mollusk`.

redmannequin:
> Sorry for the delay on reviewing here!
> 
> I like the approach of using a new method to avoid the breaking change on `execute`, however, the additional field to `MolluskComputeUnitBenchResult` is a breaking change according to Rust's type safety.
> 
> You make a good point that right now, developers can't opt-out of the file write. The default behavior is to write the file to a `benches` folder in the root, which is honestly not great. So I think one solid change is to offer that ability to opt-out of file writes. Since the fields on the `MolluskComputeUnitBencher` struct are private, we have some flexibility here. Maybe we can just add a `skip_markdown()` method to the builder?
> 
> As for returning bench results, I wonder if you are better off just writing benchmarks with `Mollusk` directly. There's not really anything you're doing in this PR that necessitates the `MolluskComputeUnitBencher` over `Mollusk`.

I guess the only point of interest I would like to raise is the partitioning of the logs, for every bench the logs of the single execution is stored in its respective `MolluskComputeUnitBenchResult`. An alternative solution would be allow a middleware though that seems a bit overkill but im happy to do it. 

```rust
MolluskComputeUnitBencher::new(mollusk)
     .bench(("bench0", &instruction0, &accounts0))
     .bench(("bench1", &instruction1, &accounts1))
     // wrap every bench run with before and after code
     .wrap(|mollusk, name| todo!("code before bench runs"), |mollusk| todo!("code after bench has ran"))
     .must_pass(true)
     .skip_markdown()
     .execute();
``` 

But you're right this kinda seems out of place for a benchmark utility. I just like the api you built and I would probably be duplicating a lot of the same code on my end.

If we the result fields are inaccessible, then there's no real reason to provide the option to disable the markdown write as there's not much benefit to running the bench without it. 

Let me know if we should close this PR.   

An no worries on the late reply, ce la vie.

buffalojoec:
> I guess the only point of interest I would like to raise is the partitioning of the logs, for every bench the logs of the single execution is stored in its respective `MolluskComputeUnitBenchResult`. An alternative solution would be allow a middleware though that seems a bit overkill but im happy to do it.

Yeah, like I said this doesn't seem like it makes sense for the benchmark tool. You can just write your benchmark tests with `Mollusk` and add all of the custom functionality you need. The benchmark tool just wraps `Mollusk` in a builder API for creating "benches", which are really just invocations of `mollusk.process_instruction` plus the markdown output for CUs.

You can accomplish what you're looking to do and beyond by just writing benchmarks with `Mollusk` itself.

PR #107, opened 2025-04-27
fix: `keyed_account_for_bpf_loader_v3_program` index

sonicfromnewyoke:
### Problem

functions `keyed_account_for_bpf_loader_v2_program()` and `keyed_account_for_bpf_loader_v3_program()` resulted in the same output due to mistyped index into `BUILTINS`

### Solution
use correct index into `BUILTINS` slice


buffalojoec (approved):
Thanks!

PR #108, opened 2025-04-28
fix: use correct license name in Cargo.toml

sonicfromnewyoke:
### Problem
`LICENSE` file in the root is **Apache-2.0**, however packages are published to the `crates.io` under the **MIT**

### Solution
use proper license name in the root `Cargo.toml`


buffalojoec (commented):
Lol, thanks! Can we actually use `license-file` ?

buffalojoec (approved):

PR #111, opened 2025-04-30
harness: add rent-exemption account check

buffalojoec:
Adds a new account check to assert a resulting account is rent-exempt.

```rust
Check::account(&key)
    .rent_exempt() // <--
    .build()
```

When the check is provided to `Mollusk::process_and_validate_instruction`, it uses the Rent instance located at `mollusk.sysvars.rent`. However, this check can also be provided to functions like `InstructionResult::run_checks`.

Unfortunately, in order to avoid a breaking change, I had to introduce a new method `InstructionResult::run_checks_with_context` which allows the Rent instance to be provided via trait implementation. the trait approach gives us much more flexibility in the future if we need to add more context items. However, the drawback is that both `InstructionResult::run_checks` and `InstructionResult::run_checks_with_config` will always use the default Rent instance (ie. `Rent::default()`).

I added documentation to reflect this potential footgun. The alternative is to make this a breaking change, and simply replace `run_checks` with this new trait-based approach, and drop the others.

Since you requested this feature @jstarry, let me know if you have thoughts.

The breaking-change alternative is here: https://github.com/anza-xyz/mollusk/pull/112

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -393,7 +393,7 @@ use {
     agave_feature_set::FeatureSet,
     agave_precompiles::get_precompile,
     mollusk_svm_error::error::{MolluskError, MolluskPanic},
-    result::Config,
+    result::{Config, DefaultCheckContext},
     solana_account::Account,
     solana_compute_budget::compute_budget::ComputeBudget,
     solana_fee_structure::FeeStructure,
@@ -682,7 +682,13 @@ impl Mollusk {
         #[cfg(any(feature = "fuzz", feature = "fuzz-fd"))]
         fuzz::generate_fixtures_from_mollusk_test(self, instruction, accounts, &result);
 
-        result.run_checks_with_config(checks, &self.config);
+        result.run_checks_with_context(
+            checks,
+            &DefaultCheckContext {
+                config: &self.config,
+                rent: &self.sysvars.rent,
+            },
+        );
         result
     }
 
--- a/harness/src/result.rs
+++ b/harness/src/result.rs
@@ -266,12 +266,20 @@ impl InstructionResult {
     }
 
     /// Perform checks on the instruction result.
+    ///
+    /// Note: Uses `Rent::default()`. If you wish to provide a custom `Rent`
+    /// instance, or the `Rent` instance on your `Mollusk` instance, use
+    /// `run_checks_with_context`.
     pub fn run_checks_with_config(&self, checks: &[Check], config: &Config) -> bool {
         let rent = &Rent::default();
         self.run_checks_with_context(checks, &DefaultCheckContext { config, rent })
     }
 
     /// Perform checks on the instruction result, panicking on any mismatches.
+    ///
+    /// Note: Uses `Rent::default()`. If you wish to provide a custom `Rent`
+    /// instance, or the `Rent` instance on your `Mollusk` instance, use
+    /// `run_checks_with_context`.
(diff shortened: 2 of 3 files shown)


jstarry:
In order to match the SVM behavior used in Agave, the rent check needs to be a little more complex:
1. Accounts with a zero lamport balance do not need to pass rent exemption checks
2. The incinerator address is exempted from rent exemption
3. Rent paying accounts are allowed to persist if they were already rent-paying (as long as their data didn't change and their balance wasn't topped up). Despite there no longer being any rent-paying accounts on any of the main clusters, some builtin accounts are technically rent paying and should not fail a transaction if rent exemption checks are enabled.

Also, I think this check should be done inside `Mollusk::process_instruction` by default rather than be added as an explicit separate check. That way you won't have to worry about the API change. But of course this would be a breaking change in terms of behavior but I think it's worthwhile to make this change despite that.

buffalojoec:
So, this tool doesn't actually deal with transactions, so there's nowhere in here that I want to throw a `TransactionError` (such as `TransactionError::InsufficientFundsForRent`). The idea is to test only the state transition function of the sBPF virtual machine, so none of the transaction pipeline exists here. For all intents and purposes, this tool's coverage domain exists from VM serialization to deserialization.

I think adding rent-exemption checks is a nice-to-have check for program devs to unit test state transitions on accounts under some program(s). Mollusk should only allow for checks against `InstructionError` (or `ProgramError`) - in this case `InstructionError::AccountNotRentExempt` - which is only thrown by programs themselves, like System.

For that reason, we probably don't need the special-casing from the runtime, however this first bullet is interesting:

> 1. Accounts with a zero lamport balance do not need to pass rent exemption checks

Does it matter who the owner is?

jstarry:
> So, this tool doesn't actually deal with transactions, so there's nowhere in here that I want to throw a TransactionError (such as TransactionError::InsufficientFundsForRent). The idea is to test only the state transition function of the sBPF virtual machine, so none of the transaction pipeline exists here. For all intents and purposes, this tool's coverage domain exists from VM serialization to deserialization.

Ah totally makes sense. Yeah doesn't make sense to enable tx level rent checks by default here in mollusk

> we probably don't need the special-casing from the runtime

I think we do because I think sysvars in mollusk will be created as rent-paying. So if anyone uses mollusk is testing their instruction which reads a ssyvar and also wants to add the rent exempt check, the rent exempt check will always fail on the sysvar. Similarly, if their instruction burns sol by sending to the incinerator, we need an exemption for the incinerator or else rent exempt checks would fail if they burn too little sol.

> > Accounts with a zero lamport balance do not need to pass rent exemption checks
> Does it matter who the owner is?

Nope. If an accounts has zero lamports after svm processing, it will be deleted no matter what the owner is.

buffalojoec:
> I think we do because I think sysvars in mollusk will be created as rent-paying. So if anyone uses mollusk is testing their instruction which reads a ssyvar and also wants to add the rent exempt check, the rent exempt check will always fail on the sysvar. Similarly, if their instruction burns sol by sending to the incinerator, we need an exemption for the incinerator or else rent exempt checks would fail if they burn too little sol.

This is all true, but we only need to special-case if we bake this in as a panic for post-execution. With the approach in this PR, it only checks the account you tell it to, so devs would just not tell it to check sysvars or incinerator for rent-exemption.

buffalojoec:
Going with the breaking change btw. Feel free to keep discussion going here though!

PR #114, opened 2025-04-30
harness: introduce account store interface and wrapper

buffalojoec:
Adds a new API for working with Mollusk using a provided account store.

A new trait - `AccountStore` - allows developers to implement their own custom account store and provide it to a new method on Mollusk: `with_account_store`. This method essentially "converts" from a `Mollusk` instance to a `MolluskWithAccountStore` instance, which offers an extremely similar API, just with the `accounts` parameter removed.

The main benefit here is that `MolluskWithAccountStore` will handle the updates to accounts in between instruction invocations. As long as an instruction succeeds, all account updates (formerly from `resulting_accounts`) will be propagated to the provided account store. Essentially, it's an on-the-fly AccountsDB.


harness/src/program.rs
```diff
@@ -40,33 +46,51 @@ pub mod precompile_keys {
 }
 
 pub struct ProgramCache {
-    cache: RwLock<ProgramCacheForTxBatch>,
+    cache: Rc<RefCell<ProgramCacheForTxBatch>>,
```
grod220, line 49: Did we lose threadsafety here? Should we go with a `Arc<RwLock<…>>`?

harness/src/lib.rs
```diff
@@ -965,4 +967,148 @@ impl Mollusk {
...
+/// provide a slice of accounts.
+///
+/// The `MolluskWithAccountStore` API internally updates the account store
+/// after each successful instruction execution, making it an ideal candidate
+/// for longer-form testing across many instructions, most likely chains.
+///
+/// Besides the omission of the `accounts` input parameter and the
+/// `resulting_accounts` return field, the API is functionally identical to the
+/// `Mollusk` API.
+pub struct MolluskWithAccountStore<AS: AccountStore> {
```
grod220: nit: maybe MolluskStore conveys the same thing (?)

harness/src/account_store.rs
```diff
@@ -0,0 +1,18 @@
...
+/// `MolluskWithAccountStore`.
+use {solana_account::Account, solana_pubkey::Pubkey};
+
+/// A trait for implementing an account store, to be used with the
+/// `MolluskWithAccountStore`.
+pub trait AccountStore {
+    /// Returns the default account to be used when an account is not found.
+    fn default_account(&self, _pubkey: &Pubkey) -> Account {
+        Account::default()
+    }
```
grod220, line 11: Should we have a fallback like this at all? It feels like this could swallow the error of origin and later the consumer may get an somewhat unrelated error message.

grod220 (commented):
I like this! Would also be great to have docs to show how the chaining works.

harness/src/account_store.rs
buffalojoec: I think we should, only because I think we should make the base behavior operate just like the Solana runtime. If an account isn't found in AccountsDB, it automatically defaults to a system account with zero lamports, zero space, zero data, etc.

If we don't, then the harness would just panic, and you'd have to go manually insert some kind of default or mock account.

harness/src/lib.rs
buffalojoec: Haha I know the name is pretty verbose, but anything more concise felt like it was misleading. `MolluskStore` for example makes me think the whole thing is an account store. I'm open to something like `MolluskContext` or `MolluskScaffold`, since we could add more features to those without obfuscating the name, but I'm not entirely sure tbh.

harness/src/program.rs
buffalojoec: Yes, we lose thread safety here, but I'm not actually making any guarantees that this library is thread-safe. I would have to comb through all of the Agave and SDK APIs it's using to make sure those are all thread-safe before making any assertions.

I made this change really just to have a faster borrow. RwLock felt like unnecessary overhead if we're not guaranteeing thread safety.

Wdyt?

harness/src/account_store.rs
grod220: Oh interesting. Yes, makes sense to match the runtime behavior. Wonder why it does that versus throwing a specific AccountNotFound error.

harness/src/lib.rs
grod220: Gotcha, think the original is ok to go with 👍

harness/src/program.rs
grod220: I think the question is: will we expect consumers to create a single Mollusk instance and share it between tests? (given cargo test uses parallel threads by default).

We haven't done this so far, is there a usecase there we are missing?

buffalojoec: Hm, good point. I suppose that would be pretty slick if you shared a Mollusk instance across parallel tests. It's an immutable reference to `&self` to call into methods like `process_and_validator_instruction`, so maybe that's the move then.

harness/src/account_store.rs
buffalojoec: Mostly so you can do things like transfer to an address that isn't "registered" with the chain yet.

harness/src/lib.rs
stegaBOB: MolluskContext would be great here IMO. With `MolluskWithAccountStore`, I'd be instantly aliasing it wherever its imported lol

harness/src/program.rs
buffalojoec: Right now, even if I flip these back to `RwLock` or use something like Arc<Mutext<..>>`, there are a few other APIs that aren't thread-safe, like the log collector. I'm inclined to worry about thread safety later, if the need arises. Wdyt?

stegaBOB: Partially biased, since thread safety isn't something I envision us needing, but this seems to be pretty internal here. Based on this PR so far it seems like adding thread safety can be done without breaking changes in the future. Based on that I'd be in favor of holding off for now

PR #133, opened 2025-06-16
update t22 to latest version

deanmlittle:
Bumped t22 as current binary is missing some of the latest updated features.


Cargo.toml
```diff
@@ -18,7 +18,7 @@ repository = "https://github.com/anza-xyz/mollusk"
 readme = "README.md"
 license-file = "LICENSE"
 edition = "2021"
-version = "0.2.0"
+version = "0.2.1"
```
buffalojoec: Don't worry about this I'll do the bumps.

```suggestion
version = "0.2.0"
```

buffalojoec: Ahh might have to do this one locally to get the lockfile change.

buffalojoec (commented):
Lgtm just skip the bump

Cargo.toml
deanmlittle: merged

PR #134, opened 2025-06-18
chore: add `rust-toolchain.toml`

sonicfromnewyoke:
### Problem
right now there is no way to auto provide `rustup` toolchains and their  associated pieces. to install, remove, update, etc

### Potential Solution
use `rust-toolchain.toml` and pin channel to `1.86.0`

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "1.86.0"
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.86.0"
+channel = "1.84.1"


rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.86.0"
```
sonicfromnewyoke: maybe it should be as `nightly-2024-11-22` for compatibility with CI ? but i would prefer to update nightly version in CI step instead of this

buffalojoec: Agreed, let's keep only the stable version in the toml.

buffalojoec: Let's go with what's in Agave 2.2.

```suggestion
channel = "1.84.1"
```

PR #135, opened 2025-06-24
Update solana-bpf-loader-program

grod220:
Needed in Token-Wrap to be able to use 2.3.1 agave deps.

In progress, some blockers with SVMFeatureSet.


harness/src/lib.rs
```diff
@@ -625,19 +632,22 @@ impl Mollusk {
...
                 self.logger.clone(),
-                self.compute_budget,
+                SVMTransactionExecutionBudget::default(),
+                Default::default(),
             );
             if let Some(precompile) = get_precompile(&instruction.program_id, |feature_id| {
-                invoke_context.get_feature_set().is_active(feature_id)
+                // invoke_context.get_feature_set().is_active(feature_id)
+                // There is not way to take a SVMFeatureSet and see if the feature pubkey
+                // matches a field
```
grod220: @buffalojoec maybe you have some ideas here. Sadly, we don't have a way to take a feature pubkey and see if it is enabled within a `SVMFeatureSet`.

buffalojoec: Yeah, this is really gross now, because we've extracted precompile management into the validator and outside of SVM. The implementation will look a little wonky, because now our callback has to access the original feature set, and it also has to provide the SVM feature set to the `InvokeContext`.

We basically have to mock this implementation of `InvokeContextCallback` on `Bank` for the Mollusk mock callback you laid down.

https://github.com/anza-xyz/agave/blob/e6824630602a76342246bfbc943a452d9661935b/runtime/src/bank.rs#L6640-L6654

buffalojoec (dismissed):

Issue #137, opened 2025-06-30
macOS linker complains at sol_burn_cus


rustopian:
```
Undefined symbols for architecture arm64:
            "_sol_burn_cus", referenced from:
                custom_syscall_program::process_instruction
```

Fix proposed in #138

PR #138, opened 2025-06-30
add no-op to satisfy linker for cargo check, test, etc.

rustopian:
No-op to fix https://github.com/anza-xyz/mollusk/issues/137


rustopian:
sgtm, I'll update

buffalojoec (approved):
Lgtm, thanks!

PR #139, opened 2025-07-02
feat: account fetchers

buffalojoec:
Adds support for a handful of account fetching libraries, which should help developers obtain `Account` instances from common sources, such as fixtures and the Solana JSON-RPC.

cc @idatsy

---

Note I've intentionally reimplemented things like `UiAccount` and RPC types to avoid any heavy dependencies from Agave.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,9 +162,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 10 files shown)


buffalojoec:
Closing for now! Seems like this is maybe not needed anymore.

PR #143, opened 2025-07-16
feat: add invoke context inspection callback

ssmike:
We are developing an instruction level tracer/emulator for svm. 

This change allows the user to reuse runtime setup logic, programs loading and result compilation while allowing to inspect low level details like registers trace and transaction context

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -452,6 +452,7 @@ pub mod sysvar;
 pub use mollusk_svm_result as result;
 #[cfg(any(feature = "fuzz", feature = "fuzz-fd"))]
 use mollusk_svm_result::Compare;
+use solana_transaction_context::{IndexOfAccount, InstructionAccount};
 use {
     crate::{
         account_store::AccountStore, compile_accounts::CompiledAccounts, epoch_stake::EpochStake,
@@ -489,6 +490,8 @@ pub struct Mollusk {
     pub program_cache: ProgramCache,
     pub sysvars: Sysvars,
 
+    pub invocation_inspect_callback: Box<dyn InvocationInspectCallback>,
+
     /// This field stores the slot only to be able to convert to and from FD
     /// fixtures and a Mollusk instance, since FD fixtures have a
     /// "slot context". However, this field is functionally irrelevant for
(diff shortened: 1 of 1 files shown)


buffalojoec (commented):
I'm not totally against this, but I'm struggling a bit to see the motivation.

The first four arguments of the before callback are just the inputs a user provides in the form of instruction accounts. If you need them in the form of `&[InstructionAccount]`, you can just use the [`keys`](https://github.com/anza-xyz/mollusk/tree/main/keys) library.

The `InvokeContext` before and after could be useful, but it's also a representation of whatever you've configured in the `Mollusk` instance, along with the same inputs as above (`TransactionContext`).

However, since it seems like you're after mostly post-invocation `InvokeContext`, and you may want to compare that to the `InvokeContext` before the instruction was invoked, what about just keeping this callback to one argument on both sides: `InvokeContext`?

ssmike:
Yes, these are easily derived from user instruction, but accounts should be consistent with transaction context. The keymap/compile_accounts from keys are unstable due to hashmaps used and so are unusable here. I added these arguments for convenience, but in case you are against it I don't mind implementing matching logic outside.

Also in case we are running instruction chain, it would be convenient to differentiate somehow which instruction it is.

harness/src/lib.rs
```diff
@@ -489,6 +490,8 @@ pub struct Mollusk {
     pub program_cache: ProgramCache,
     pub sysvars: Sysvars,
 
+    pub invocation_inspect_callback: Box<dyn InvocationInspectCallback>,
```
buffalojoec, line 498: Can you add a doc comment to this?

Also, I think this field should exist behind a feature flag. It can behave just like the `fuzz-fd` feature-flagged `slot` field below it. Let's call it `invocation-inspect-callback`.

harness/src/lib.rs
```diff
@@ -498,6 +501,35 @@ pub struct Mollusk {
...
 
+pub trait InvocationInspectCallback {
+    fn before_invocation(
+        &self,
+        program_id: &Pubkey,
+        instruction_data: &[u8],
+        instruction_accounts: &[InstructionAccount],
+        program: IndexOfAccount,
+        invoke_context: &InvokeContext,
+    );
```
buffalojoec:
```suggestion
    fn before_invocation(
        &self,
        program_id: &Pubkey,
        program_index: IndexOfAccount,
        instruction_accounts: &[InstructionAccount],
        instruction_data: &[u8],
        invoke_context: &InvokeContext,
    );
```

harness/src/lib.rs
```diff
@@ -498,6 +501,35 @@ pub struct Mollusk {
...
+
+    fn after_invocation(&self, invoke_context: &InvokeContext);
+}
+
+pub struct EmptyInvocationInspectCallback {}
```
buffalojoec: Does this need to be `pub`?

```suggestion
struct EmptyInvocationInspectCallback;
```

ssmike: Yes I think it should be public behind the same feature flag so that user can construct a mollusk instance without imlementing the trait. Otherwise, the only way to do it is to use Mollusk::default or Mollusk::new(pubkey, filename)

buffalojoec (approved):
Looks good!

PR #144, opened 2025-07-16
fix: add token-2022 program with loader v3

dhl:
Use loader v3 to add token-2022 to align with onchain deployment


sonicfromnewyoke (commented):
good catch, LGTM

buffalojoec (approved):
Thanks!

PR #145, opened 2025-07-25
Fix Rent exempt check

deanmlittle:
Currently in Mollusk, when we run a `Check::rent_exempt` or `Check::all_rent_exempt` on an execution result, it will return a false negative for accounts that belong to `SystemProgram` with a lamport balance of `0`, such as uninitialized PDAs that are used only for signing purposes. In a regular validator environment, such accounts would simply be cleaned up and would not result in an error. In order to replicate this functionality in Mollusk, I have modified the rent exempt checks to not return an error in such cases.

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -577,8 +577,8 @@ impl Default for Mollusk {
 }
 
 impl CheckContext for Mollusk {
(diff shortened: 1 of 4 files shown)


buffalojoec (approved):
Thanks!! 🙏 g2g from my side once the test is fixed.

PR #146, opened 2025-07-31
Feat: added state helpers for the token program

L0STE:


programs/token/src/state.rs
```diff
@@ -0,0 +1,211 @@
...
+        owner: token_program.unwrap_or(token::ID),
+        executable: false,
+        rent_epoch: 0,
+    };
+
+    (pubkey.unwrap_or(Pubkey::new_unique()), account)
+}
+
+// Create a Keyed Account for a Token Account with default data
+#[allow(dead_code)]
```
sonicfromnewyoke: why do you need this?

programs/token/src/state.rs
sonicfromnewyoke, line 1: not sure that `state.rs` is a good naming for it.

what do you think about keeping those helpers inside the corresponding program files?
like ` fn keyed_account_for_associated_token_account(...)` inside `programs/src/associated_token.rs`

the same with  `fn keyed_account_for_mint(...)`  and `fn keyed_account_for_token_account(...)` in `token.rs`

and `token2022.rs` will contain setup with extensions

programs/token/Cargo.toml
```diff
@@ -19,3 +19,5 @@ token-2022 = []
 mollusk-svm = { workspace = true }
 solana-account = { workspace = true }
 solana-pubkey = { workspace = true }
+spl-token = "8.0.0"
+spl-associated-token-account = "7.0.0"
```
buffalojoec: Can you declare these versions in the workspace manifest?

programs/token/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 //! Last updated at mainnet-beta slot height: 347196212
+mod state;
```
buffalojoec: Isn't this supposed to be public?

programs/token/src/state.rs
buffalojoec: I agree! I think we should collocate each state type with its respective program module.

programs/token/src/state.rs
```diff
@@ -0,0 +1,211 @@
...
+
+// Create a Keyed Account for a Mint with default data
+#[allow(dead_code)]
+pub fn keyed_account_for_mint_default(
+    mollusk: &Mollusk,
+    authority: &Pubkey,
+    decimals: u8,
+    pubkey: Option<Pubkey>,
+    token_program: Option<Pubkey>,
+) -> (Pubkey, Account) {
```
buffalojoec: I wonder if it's better to just accept a `Mint` struct in this function. This way, you get full control over the mint state customization, but the helper does the annoying part of serializing it into an account and setting up the lamports, space and owner. What do you think?

buffalojoec: We should also just have separate helpers for Token-2022, rather than a `token_program` parameter. I say this because introducing extensions bungles this whole API and makes it unintuitive, which is kind of what happened with the SPL Token CLI. Instead, we could have a separate "keyed account" API for each program, and the Token-2022 one can accept Token-2022-specific stuff like extensions.

programs/token/src/state.rs
```diff
@@ -0,0 +1,211 @@
...
+
+/// Note: Support for Token2022 state with extensions is not yet implemented
+
+// Create a Keyed Account for a Mint with default data
+#[allow(dead_code)]
+pub fn keyed_account_for_mint_default(
+    mollusk: &Mollusk,
+    authority: &Pubkey,
+    decimals: u8,
+    pubkey: Option<Pubkey>,
```
buffalojoec: I think this pattern of an optional pubkey that defaults to `Pubkey::new_unique()` isn't that useful. I think this whole API can actually just serve an "account".

Most test suites are going to do something like this:

```rust
// Keys
let mint_address = Pubkey::new_unique();
let authority_address = Pubkey::new_unique();
/* ... */

// Accounts
let mint_account = create_account_for_mint(mint_state);
let authority_account = Account::default();
/* ... */

mollusk.process_instruction(
    &ix,
    &[
        (mint_address, mint_account),
        (authority_address, authority_account),
        /* ... */
    ],
);
```

Or just this:

```rust
mollusk.process_instruction(
    &ix,
    &[
        (Pubkey::new_unique(), create_account_for_mint(mint_state)),
        (Pubkey::new_unique(), Account::default()),
        /* ... */
    ],
);
```

buffalojoec (commented):
Thanks for the PR @L0STE! I think we can make a few adjustments and build it out more to get this in.

I'm not worried about Token-2022 right now, so if you don't need state helpers for that program, we can wait on that. Let's just get SPL-Token and ATA worked into the existing modules and we can ship.

programs/token/src/state.rs
L0STE: Yeah I think you're right! And people could still alway fill it up with using Default if they want to only supply certain fields.

programs/token/src/state.rs
L0STE: I agree, I left the (Pubkey, Account) API only for the `keyed_account_for_associated_token_account` helpers so people doesn't need to include the `spl-associated-token` crate when they want to create Associated Token Accounts.

programs/token/Cargo.toml
```diff
@@ -19,3 +19,5 @@ token-2022 = []
 mollusk-svm = { workspace = true }
 solana-account = { workspace = true }
 solana-pubkey = { workspace = true }
+spl-token = { workspace = true }
+spl-associated-token-account = { workspace = true }
```
sonicfromnewyoke:
```suggestion
spl-token = { workspace = true, optional = true  }
spl-associated-token-account = { workspace = true, optional = true  }
```

programs/token/src/state.rs
buffalojoec: Cool, that makes sense, and it's a nice touch!

programs/token/Cargo.toml
sonicfromnewyoke: with some change in `[features]` like follows:

```toml
[features]
# ...
token = ["dep:spl-token"]
associated-token = ["dep:spl-associated-token-account "]
```

should give a better control over the fetched deps by feature

programs/token/src/associated_token.rs
```diff
@@ -22,3 +28,53 @@ pub fn account() -> Account {
 pub fn keyed_account() -> (Pubkey, Account) {
     (ID, account())
 }
+
+/// Keyed Account for an Associated Token Account
+pub fn keyed_account_for_associated_token_account(
+    mollusk: &Mollusk,
```
buffalojoec: No need for an entire `&Mollusk` reference when all we need is `&Rent`. In fact, if you want, we can just eliminate this completely and use `Rent::default()` within the helper, since I already (perhaps mistakenly) introduced this pattern in the harness's program module.

https://github.com/anza-xyz/mollusk/blob/e50cf9fbc69b92f347978bb5528192d90d8a88a3/harness/src/program.rs#L303

I'll leave it up to you!

L0STE:
Agreed with all the comments! I initially based this on our internal implementation, but the suggested approach is much better.

A few implementation notes:
- I kept the `Mollusk` parameter for account creation, though it's only used for rent calculations. We could simplify this by using the `Rent` sysvar directly instead but I personally think that the current implementation is fine.
- I'm returning `(Pubkey, Account)` for associated token account creation to save developers from having to import the `spl-associated-token` crate just to derive the address.
- I've added basic Token2022 account support. When we're ready to handle extensions, we can add a `keyed_account_for_mint_with_extension` helper method.

programs/token/Cargo.toml
buffalojoec: I had literally the same comment queued up. Nice!

programs/token/src/token.rs
```diff
@@ -22,3 +28,34 @@ pub fn account() -> Account {
 pub fn keyed_account() -> (Pubkey, Account) {
     (ID, account())
 }
+
+/// Create a Keyed Account for a Mint
+pub fn keyed_account_for_mint(mollusk: &Mollusk, mint_data: Mint) -> Account {
```
buffalojoec:
```suggestion
pub fn create_account_for_mint(mollusk: &Mollusk, mint_data: Mint) -> Account {
```

programs/token/src/token.rs
```diff
@@ -22,3 +28,34 @@ pub fn account() -> Account {
...
+        lamports: mollusk.sysvars.rent.minimum_balance(Mint::LEN),
+        data,
+        owner: ID,
+        executable: false,
+        rent_epoch: 0,
+    }
+}
+
+/// Create a Keyed Account for a Token Account
+pub fn keyed_account_for_token_account(
```
buffalojoec:
```suggestion
pub fn create_account_for_token_account(
```

programs/token/src/token2022.rs
```diff
@@ -22,3 +22,35 @@ pub fn account() -> Account {
 pub fn keyed_account() -> (Pubkey, Account) {
     (ID, account())
 }
+
+/// Create a Keyed Account for a Mint
+pub fn keyed_account_for_mint(mollusk: &Mollusk, mint_data: Mint) -> Account {
```
buffalojoec:
```suggestion
pub fn create_account_for_mint(mollusk: &Mollusk, mint_data: Mint) -> Account {
```

programs/token/src/token2022.rs
```diff
@@ -22,3 +22,35 @@ pub fn account() -> Account {
...
+        lamports: mollusk.sysvars.rent.minimum_balance(Mint::LEN),
+        data,
+        owner: ID,
+        executable: false,
+        rent_epoch: 0,
+    }
+}
+
+/// Create a Keyed Account for a Token Account
+pub fn keyed_account_for_token_account(
```
buffalojoec:
```suggestion
pub fn create_account_for_token_account(
```

programs/token/src/associated_token.rs
sonicfromnewyoke: i believe @L0STE  implementation is slightly more robust because we intentionally use configuration for the rent sysvar from the runtime.

`Rent::default()` will return a different config than what was configured for the test case, isn't it?

buffalojoec: Yes that's why I said it was probably a mistake to do it in the place I linked. I still think you should only make the parameter require what you need, ie. `&Rent`.

PR #147, opened 2025-08-09
fixes typo

ANISH-SR:
"the" -> "to" 

david says I got a sharp eye : )
image


buffalojoec (approved):
> david says I got a sharp eye : )

Not too shabby!

PR #155, opened 2025-08-26
Use 3.0 crates and 2.0 SPL interfaces

rustopian:
Updates Mollusk to use 3.0 crates (and 2.0 SPL interface crates when available).

### DIFF
--- a/fuzz/fixture-fd/src/context.rs
+++ b/fuzz/fixture-fd/src/context.rs
@@ -54,7 +54,7 @@ impl From<EpochContext> for ProtoEpochContext {
 }
 
 /// Instruction context fixture.
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Debug, Default)]
 pub struct Context {
     /// The program ID of the program being invoked.
     pub program_id: Pubkey,
--- a/fuzz/fixture-fd/src/lib.rs
+++ b/fuzz/fixture-fd/src/lib.rs
@@ -29,7 +29,7 @@ use {
 
 /// A fixture for invoking a single instruction against a simulated SVM
 /// program runtime environment, for a given program.
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Debug, Default)]
 pub struct Fixture {
     /// The fixture metadata.
     pub metadata: Option<Metadata>,
@@ -154,13 +154,7 @@ mod tests {
         let instruction_accounts = accounts
             .iter()
             .enumerate()
-            .map(|(i, _)| InstructionAccount {
-                index_in_transaction: i as u16,
-                index_in_caller: i as u16,
-                index_in_callee: i as u16,
-                is_signer: false,
-                is_writable: true,
-            })
+            .map(|(i, _)| InstructionAccount::new(i as u16, false, true))
             .collect::<Vec<_>>();
         let instruction_data = vec![4; 24];
         let slot_context = SlotContext { slot: 42 };
--- a/fuzz/fixture/proto/compute_budget.proto
+++ b/fuzz/fixture/proto/compute_budget.proto
@@ -15,7 +15,6 @@ message ComputeBudget {
     uint64 max_call_depth = 10;
     uint64 stack_frame_size = 11;
     uint64 log_pubkey_units = 12;
-    uint64 max_cpi_instruction_size = 13;
     uint64 cpi_bytes_per_unit = 14;
     uint64 sysvar_base_cost = 15;
     uint64 secp256k1_recover_cost = 16;
--- a/fuzz/fixture/src/context.rs
+++ b/fuzz/fixture/src/context.rs
@@ -14,7 +14,7 @@ use {
 };
 
 /// Instruction context fixture.
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct Context {
     /// The compute budget to use for the simulation.
     pub compute_budget: ComputeBudget,
@@ -30,6 +30,20 @@ pub struct Context {
     pub instruction_data: Vec<u8>,
     /// Input accounts with state.
     pub accounts: Vec<(Pubkey, Account)>,
+}
+
+impl Default for Context {
+    fn default() -> Self {
+        Self {
+            compute_budget: ComputeBudget::new_with_defaults(true),
+            feature_set: FeatureSet::default(),
+            sysvars: Sysvars::default(),
+            program_id: Pubkey::default(),
+            instruction_accounts: Vec::default(),
+            instruction_data: Vec::default(),
+            accounts: Vec::default(),
+        }
+    }
 }
 
 impl From<ProtoContext> for Context {
@@ -64,7 +78,10 @@ impl From<ProtoContext> for Context {
             .collect();
 
         Self {
-            compute_budget: value.compute_budget.map(Into::into).unwrap_or_default(),
+            compute_budget: value
+                .compute_budget
+                .map(Into::into)
+                .unwrap_or_else(|| ComputeBudget::new_with_defaults(true)),
             feature_set: value.feature_set.map(Into::into).unwrap_or_default(),
             sysvars: value.sysvars.map(Into::into).unwrap_or_default(),
             program_id,
--- a/harness/Cargo.toml
+++ b/harness/Cargo.toml
@@ -10,7 +10,7 @@ edition = { workspace = true }
 version = { workspace = true }
 
 [features]
-default = []
+default = ["serde"]
 all-builtins = [
     "dep:solana-stake-program",
 ]
@@ -32,6 +32,7 @@ invocation-inspect-callback = []
 [dependencies]
 agave-feature-set = { workspace = true }
 agave-precompiles = { workspace = true }
+agave-syscalls = { workspace = true }
 bincode = { workspace = true }
 serde = { workspace = true, features = ["derive"], optional = true }
 mollusk-svm-error = { workspace = true }
@@ -48,9 +49,9 @@ solana-epoch-rewards = { workspace = true }
 solana-epoch-schedule = { workspace = true }
 solana-hash = { workspace = true }
 solana-instruction = { workspace = true }
+solana-instruction-error = { workspace = true, features = ["serde"] }
 solana-loader-v3-interface = { workspace = true, features = ["serde"] }
 solana-loader-v4-interface = { workspace = true }
-solana-log-collector = { workspace = true }
 solana-logger = { workspace = true }
 solana-precompile-error = { workspace = true }
 solana-program-error = { workspace = true }
@@ -58,14 +59,16 @@ solana-program-runtime = { workspace = true }
 solana-pubkey = { workspace = true }
 solana-rent = { workspace = true }
 solana-sdk-ids = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
 solana-slot-hashes = { workspace = true }
 solana-stake-interface = { workspace = true }
 solana-stake-program = { workspace = true, optional = true }
 solana-svm-callback = { workspace = true }
+solana-svm-log-collector = { workspace = true }
+solana-svm-timings = { workspace = true }
 solana-system-program = { workspace = true }
 solana-sysvar = { workspace = true }
 solana-sysvar-id = { workspace = true }
-solana-timings = { workspace = true }
 solana-transaction-context = { workspace = true, features = ["dev-context-only-utils"] }
 
 [dev-dependencies]
--- a/harness/src/sysvar.rs
+++ b/harness/src/sysvar.rs
@@ -58,7 +58,10 @@ impl Default for Sysvars {
 }
 
 impl Sysvars {
-    fn sysvar_account<T: SysvarId + Sysvar>(&self, sysvar: &T) -> (Pubkey, Account) {
+    fn sysvar_account<T: SysvarId + Sysvar + serde::Serialize>(
+        &self,
+        sysvar: &T,
+    ) -> (Pubkey, Account) {
         let data = bincode::serialize::<T>(sysvar).unwrap();
         let space = data.len();
         let lamports = self.rent.minimum_balance(space);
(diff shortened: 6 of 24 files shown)


harness/tests/bpf_program.rs
```diff
@@ -12,6 +12,27 @@ use {
     solana_system_interface::error::SystemError,
 };
 
+// TODO: Investigate these differences
+#[cfg(target_os = "linux")]
+const CU_WRITE_DATA: u64 = 387;
+#[cfg(not(target_os = "linux"))]
+const CU_WRITE_DATA: u64 = 384;
```
buffalojoec: Interesting. Are you sure it's the same platform-tools on both machines? Maybe the one in the Makefile needs to be adjusted?

rustopian: I suspect platform-tools too; the CU differences look the same as gains with the recent platform-tools versions. Can finish investigating this very soon.

rustopian: Makefile updated and all matches

fuzz/fixture/Cargo.toml
```diff
@@ -8,6 +8,10 @@ license-file ={ workspace = true }
 edition = { workspace = true }
 version = { workspace = true }
 
+[features]
+simd-0296 = []
```
rustopian: @buffalojoec let me know if you think this should be handled another way.

Gating [SIMD-0296](https://github.com/solana-foundation/solana-improvement-documents/pull/296) behind a Cargo feature seems best for now.

`ComputeBudget::{Default, From<_>}` can’t consult a `FeatureSet`, but just hardcoding `true`/`false` is brittle. I could introduce a `from_feature_set(&FeatureSet)` path, but that feels too invasive/overkill at this point.

The feature is added in [this commit](https://github.com/anza-xyz/mollusk/pull/155/commits/b4856ec10ddbac4db390c95b7b311b9f52013791).

fuzz/fixture-fd/src/context.rs
```diff
@@ -54,7 +54,7 @@ impl From<EpochContext> for ProtoEpochContext {
 }
 
 /// Instruction context fixture.
-#[derive(Clone, Debug, Default, PartialEq)]
```
buffalojoec, line 57: Why did we lose the `PartialEq` here?

fuzz/fixture-fd/src/lib.rs
```diff
@@ -29,7 +29,7 @@ use {
 
 /// A fixture for invoking a single instruction against a simulated SVM
 /// program runtime environment, for a given program.
-#[derive(Clone, Debug, Default, PartialEq)]
```
buffalojoec, line 32: Same here - is it because of `InstructionAccount`?

fuzz/fixture/Cargo.toml
buffalojoec: Shoot, I just realized I wrote the wrong SIMD number in there, lol. It's supposed to be [SIMD-0268](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0268-raise-cpi-nesting-limit.md).

fuzz/fixture-fd/Cargo.toml
```diff
@@ -17,6 +17,7 @@ serde_json = { workspace = true }
 solana-account = { workspace = true }
 solana-keccak-hasher = { workspace = true }
 solana-pubkey = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
```
buffalojoec: Is this used in this crate?

fuzz/fixture/proto/compute_budget.proto
```diff
@@ -15,7 +15,6 @@ message ComputeBudget {
     uint64 max_call_depth = 10;
     uint64 stack_frame_size = 11;
     uint64 log_pubkey_units = 12;
-    uint64 max_cpi_instruction_size = 13;
```
buffalojoec, line 18: Unfortunately I believe all of the other indices need to be updated :(

fuzz/fixture/src/context.rs
```diff
@@ -32,6 +32,25 @@ pub struct Context {
     pub accounts: Vec<(Pubkey, Account)>,
 }
 
+impl Default for Context {
```
buffalojoec: Actually, looking at the codebase, if I just remove the `Default` trait from `Fixture` and `Context`, everything still compiles! I wonder if we even need `Default` at all for any of this stuff.

Since this looks like it will already be a breaking change, I'm ok with removing support for `Default`, as long as it doesn't break anything.

harness/src/sysvar.rs
```diff
@@ -58,7 +58,7 @@ impl Default for Sysvars {
 }
 
 impl Sysvars {
-    fn sysvar_account<T: SysvarId + Sysvar>(&self, sysvar: &T) -> (Pubkey, Account) {
+    fn sysvar_account<T: SysvarId + SysvarSerialize>(&self, sysvar: &T) -> (Pubkey, Account) {
```
buffalojoec: I believe `SysvarSerialize` includes `SysvarId`.

harness/tests/bpf_program.rs
```diff
@@ -12,6 +12,11 @@ use {
     solana_system_interface::error::SystemError,
 };
 
+const CU_WRITE_DATA: u64 = 384;
+const CU_TRANSFER: u64 = 2533;
+const CU_CLOSE_ACCOUNT: u64 = 2608;
+const CU_CPI: u64 = 2418;
```
buffalojoec: No need for the comments up here IMO, since they're only used in one spot! Plus they are only... semi-constant :)

harness/tests/fd_test_vectors.rs
```diff
@@ -152,12 +152,19 @@ fn compare_instruction_accounts(a: &[InstructionAccount], b: &[InstructionAccoun
     let mut a_sorted = a.to_vec();
     let mut b_sorted = b.to_vec();
 
-    // Sort by Pubkey
+    // Sort by index_in_transaction
```
buffalojoec, line 155: Ha! So this comment was just wrong. Thanks!

harness/tests/fd_test_vectors.rs
```diff
@@ -152,12 +152,19 @@ fn compare_instruction_accounts(a: &[InstructionAccount], b: &[InstructionAccoun
...
-    a_sorted == b_sorted
+    // Compare sorted lists; InstructionAccount no longer implements PartialEq
+    a_sorted
+        .iter()
+        .zip(b_sorted.iter())
+        .all(|(a_item, b_item)| {
+            a_item.index_in_transaction == b_item.index_in_transaction
+                && a_item.is_signer() == b_item.is_signer()
+                && a_item.is_writable() == b_item.is_writable()
+        })
```
buffalojoec, line 167: This looks like the only place we had to do this by hand? Fine by me if we lose `PartialEq` on the fixture side, then, but only if it doesn't break the CLI.

harness/tests/precompile.rs
```diff
@@ -32,11 +47,24 @@ fn test_secp256k1() {
 #[allow(deprecated)]
```
buffalojoec: I think if you cut the deprecated helpers, we can probably lose this attribute.

buffalojoec (commented):
Cool, thanks for putting this together! Just some comments about a few things.

fuzz/fixture-fd/src/lib.rs
rustopian: Indeed, because of InstructionAccount.

rustopian: Indeed, due to `InstructionAccount`, in [7537](https://github.com/anza-xyz/agave/pull/7537). "Remove unnecessary derives that are incompatible with unsafe cell."

harness/tests/precompile.rs
rustopian: Good catch, also done in 17c5363

harness/tests/fd_test_vectors.rs
rustopian: Yes, this is the only place.

I'll see if any test needs to be added to CLI which demonstrates this breaking or not.

harness/tests/bpf_program.rs
rustopian: Fine by me, removed in 17c5363

harness/src/sysvar.rs
rustopian: Confirmed, removed in 17c5363

fuzz/fixture/Cargo.toml
rustopian: Ah, makes sense. Updated the SIMD number (in the cargo feature name) in 17c5363 and created issue [7755](https://github.com/anza-xyz/agave/issues/7755) in agave so it's not forgotten.

fuzz/fixture/src/context.rs
rustopian: Fine by me, removed in 17c5363

fuzz/fixture/proto/compute_budget.proto
rustopian: Fixed in 17c5363

fuzz/fixture-fd/Cargo.toml
rustopian: Ah, I think it was used in a workaround in development while not all 3.0 crates were available yet. Removed in 17c5363

fuzz/fixture-fd/src/context.rs
rustopian: `InstructionAccount` has lost it. Cf. https://github.com/anza-xyz/agave/pull/7537

fuzz/fixture/src/context.rs
```diff
@@ -84,11 +65,11 @@ impl From<ProtoContext> for Context {
...
         Self {
             compute_budget: value.compute_budget.map(Into::into).unwrap_or_else(|| {
-                #[cfg(feature = "simd-0296")]
+                #[cfg(feature = "simd-0268")]
                 {
                     ComputeBudget::new_with_defaults(true)
                 }
-                #[cfg(not(feature = "simd-0296"))]
+                #[cfg(not(feature = "simd-0268"))]
                 {
```
buffalojoec: Ah, I see the problem. So this should be the only spot we need this feature flag, right? 

Since these conversions happen when the fuzz API reads from a fixture, we really need to use the value in the feature set to flip this boolean.

I think you can just evaluate the feature set first [here](https://github.com/anza-xyz/mollusk/blob/1cfdd642b3afa351068148d008c0b4f066ed09c6/fuzz/fixture/src/context.rs#L67-L68), and then look up whether or not the feature is active to plumb the boolean through the compute budget conversion. We probably have to move off of `From` for `ComputeBudget` and use a free function. That's no biggie.

rustopian: Makes sense, free function incoming.

rustopian: Turns out we can easily do this without moving away from `From`. Cargo feature flag removed in dc7bb58. Also added a few tests that this works as intended in 1a50508.

fuzz/fixture/Cargo.toml
```diff
@@ -8,6 +8,9 @@ license-file ={ workspace = true }
 edition = { workspace = true }
 version = { workspace = true }
 
+[features]
+default = []
+
```
buffalojoec: No reason to add this, since there aren't any features for this crate.

fuzz/fixture/src/context.rs
```diff
@@ -63,9 +63,18 @@ impl From<ProtoContext> for Context {
...
             .collect();
 
+        let feature_set: FeatureSet = value.feature_set.map(Into::into).unwrap_or_default();
+        let simd_0268_active =
+            feature_set.is_active(&agave_feature_set::raise_cpi_nesting_limit_to_8::id());
+
+        let compute_budget = value
+            .compute_budget
+            .map(Into::into)
+            .unwrap_or_else(|| ComputeBudget::new_with_defaults(simd_0268_active));
```
buffalojoec, line 73: Nice, perfect!

fuzz/fixture/src/context.rs
```diff
@@ -133,3 +142,71 @@ pub(crate) fn hash_proto_context(hasher: &mut Hasher, context: &ProtoContext) {
     hasher.hash(&context.data);
     crate::account::hash_proto_accounts(hasher, &context.accounts);
 }
+
+#[cfg(test)]
+mod tests {
```
buffalojoec, line 147: gud tests

fuzz/fixture/Cargo.toml
```diff
@@ -21,6 +24,7 @@ solana-epoch-schedule = { workspace = true }
 solana-hash = { workspace = true }
 solana-instruction = { workspace = true }
 solana-keccak-hasher = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
```
buffalojoec: Is this used?

harness/Cargo.toml
```diff
@@ -48,24 +49,26 @@ solana-epoch-rewards = { workspace = true }
...
 solana-loader-v4-interface = { workspace = true }
-solana-log-collector = { workspace = true }
 solana-logger = { workspace = true }
 solana-precompile-error = { workspace = true }
 solana-program-error = { workspace = true }
 solana-program-runtime = { workspace = true }
 solana-pubkey = { workspace = true }
 solana-rent = { workspace = true }
 solana-sdk-ids = { workspace = true }
+solana-sha256-hasher = { workspace = true, features = ["sha2"] }
```
buffalojoec: Is this used?

fuzz/fixture/Cargo.toml
rustopian: Artifact of that simd feature, removed in 4dbabcb

fuzz/fixture/Cargo.toml
rustopian: No longer, removed in 4dbabcb

harness/Cargo.toml
rustopian: Same, removed in 4dbabcb, from here and from top level Cargo.toml

Issue #100, opened 2025-03-22
Add an helper to create the `instruction_sysvar` account


L0STE:
Due to the nature of this sysvar, it's very hard to create an implementation that resemble the one [currently used](https://github.com/anza-xyz/mollusk/blob/main/fuzz/fixture/src/sysvars.rs#L23). 

I don't have a "good" way of doing it at the moment, but this is the way that I currently do it:

```rust
//...
let (instruction_sysvar_pubkey, instruction_sysvar_account) = get_account_instructions_sysvar(&mut mollusk, &[]);
//..

fn get_account_instructions_sysvar(mollusk: &mut Mollusk, instructions: &[Instruction]) -> (Pubkey, AccountSharedData) {
    // Construct the instructions data from all instructions
    let mut data = construct_instructions_data(
        instructions.iter().map(|instruction| {
            BorrowedInstruction {
                program_id: &instruction.program_id,
                accounts: instruction.accounts
                    .iter()
                    .map(|meta| BorrowedAccountMeta {
                        pubkey: &meta.pubkey,
                        is_signer: meta.is_signer,
                        is_writable: meta.is_writable,
                    })
                    .collect(),
                data: &instruction.data,
            }
        }).collect::<Vec<_>>().as_slice()
    );

    // Find which instruction contains the sysvar account and at what position
    if let Some((ix_index, _)) = instructions.iter().enumerate()
        .find(|(_, instruction)| {
            instruction.accounts.iter()
                .any(|meta| meta.pubkey == INSTRUCTION_SYSVAR)
        }) 
    {
        store_current_index(&mut data, ix_index as u16);
    }

    // Set the account data
    let mut instruction_sysvar_account = AccountSharedData::new(
        mollusk
            .sysvars
            .rent
            .minimum_balance(data.len()),
        data.len(),
        &self::ID
    );
    instruction_sysvar_account.set_data_from_slice(&data);

    (INSTRUCTION_SYSVAR, instruction_sysvar_account)
}
```

buffalojoec:
@L0STE thanks for creating this issue! I think your method looks pretty good, but you're right we don't want to stray too far away from the implementation in the actual runtime. However, different validator clients will implement this slightly differently, so as long as the result is the same, we can mock it out in the test environment.

I'll need a little bit of time to dig into this, but I don't see why we wouldn't go at least in the general direction you've suggested here. Nice work!!

PR #156, opened 2025-09-02
feat: instructions sysvar

Arrowana:
Inspired by #100, roughly how i made it work in my fork

Note, there is the need to be able to set `top_level_instruction_index` on `TransactionContext`, otherwise the instruction in a batch are getting the wrong index, this is caused by the mollusk hack of not offering close to the original VM running and entire "transaction".

`TransactionContext::push` will set the instructions sysvar current index to `top_level_instruction_index`
https://github.com/anza-xyz/agave/blob/85d8b5859ebe462236fd22acdeaa556d82ad4f54/transaction-context/src/lib.rs#L446-L449

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2226,7 +2226,7 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm"
-version = "0.7.2"
+version = "0.8.0"
 dependencies = [
  "agave-feature-set",
  "agave-precompiles",
@@ -2286,7 +2286,7 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm-bencher"
-version = "0.7.2"
+version = "0.8.0"
 dependencies = [
  "chrono",
  "mollusk-svm",
@@ -2300,7 +2300,7 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm-cli"
-version = "0.7.2"
+version = "0.8.0"
 dependencies = [
  "chrono",
  "clap",
@@ -2319,7 +2319,7 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm-error"
-version = "0.7.2"
+version = "0.8.0"
 dependencies = [
  "solana-pubkey 4.0.0",
  "thiserror 1.0.69",
@@ -2327,7 +2327,7 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm-fuzz-fixture"
-version = "0.7.2"
+version = "0.8.0"
 dependencies = [
  "agave-feature-set",
  "mollusk-svm-fuzz-fs",
@@ -2351,7 +2351,7 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm-fuzz-fixture-firedancer"
-version = "0.7.2"
+version = "0.8.0"
 dependencies = [
  "agave-feature-set",
  "mollusk-svm-fuzz-fs",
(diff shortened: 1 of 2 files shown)


buffalojoec (commented):
Yeah I think this mockup looks good to me. Then the new module is just the logic from #100?

Arrowana:
> Yeah I think this mockup looks good to me. Then the new module is just the logic from #100?

yes, i forced pushed since i forgot the file...

Arrowana:
> > Since this is a sysvar though should this happen automagically inside mollusk setup? Rather than being an account people must add
> 
> Yep I think so! Typically I try to avoid overwriting user-provided accounts, even if it's a special account like this one. So we should make sure we first check if the account was provided already, and if it was not, then do this setup logic.
> 
> That being said, the harness does one allocation for the provided accounts slice when it creates a list of "transaction accounts" from the account compilation step.
> 
> https://github.com/anza-xyz/mollusk/blob/1cfdd642b3afa351068148d008c0b4f066ed09c6/harness/src/lib.rs#L683-L687
> 
> So I'm thinking maybe we can push it into the accounts list here? This way it's in the transaction accounts list before we even create the `TransactionContext` and do the `set_top_level_instruction_index` step.
> 
> I'm open to whether or not we should include the auto-populated Instructions Sysvar account in the resulting accounts list. I can go either way: return it or cut it out.

It can't be after `compile_accounts` because it will err with `MissingAccount` then

Arrowana:
> > Since this is a sysvar though should this happen automagically inside mollusk setup? Rather than being an account people must add
> 
> Yep I think so! Typically I try to avoid overwriting user-provided accounts, even if it's a special account like this one. So we should make sure we first check if the account was provided already, and if it was not, then do this setup logic.
> 
> That being said, the harness does one allocation for the provided accounts slice when it creates a list of "transaction accounts" from the account compilation step.
> 
> https://github.com/anza-xyz/mollusk/blob/1cfdd642b3afa351068148d008c0b4f066ed09c6/harness/src/lib.rs#L683-L687
> 
> So I'm thinking maybe we can push it into the accounts list here? This way it's in the transaction accounts list before we even create the `TransactionContext` and do the `set_top_level_instruction_index` step.
> 
> I'm open to whether or not we should include the auto-populated Instructions Sysvar account in the resulting accounts list. I can go either way: return it or cut it out.

It can't be that deep because the `process_instruction` function only knows about the current instruction...

buffalojoec:
> It can't be that deep because the `process_instruction` function only knows about the current instruction...

Ah yes, you're right...

How about this? I set up a PR that should decouple things enough for you to stick this step right before the account compilation step and use a chained iterator to feed `compile_accounts`. We also need the index setting step as well, and I'll review the Agave PR asap.

#157 

Let me know if this works for you!

quazzuk:
Any update here?

Would great to get this functionality merged.

Thanks!

buffalojoec:
Hey thanks for putting this together @Arrowana! I took this bad boy over and implemented full support for the Instructions sysvar. If you feel like reviewing, go ahead!

harness/src/lib.rs
```diff
@@ -927,15 +944,23 @@ impl Mollusk {
         accounts: &[(Pubkey, Account)],
         checks: &[Check],
     ) -> InstructionResult {
+        const INDEX: usize = 0;
```
sonicfromnewyoke, line 947: maybe it worth renaming it to something like `FIRST_INSTRUCTION_INDEX` or `TOP_LEVEL_INSTRUCTION_INDEX`, move  it to the top level of the current file, and reuse in both `process_instruction` & `process_and_validate_instruction` functions, what do you think?

buffalojoec: Is there a major benefit? It's already in-lined in the two functions. Seems good enough for me!

PR #161, opened 2025-09-30
Feature Matrix: derive baseline from fixtures, add --feature/--matrix (cartesian), reports

### LINKED ISSUE #160: Add Feature-activation Matrix runner (gated) to validate FeatureSet changes
Add a gated “Feature Matrix” to Mollusk to run the same fixtures under a baseline and one or more feature-enabled variants, compare results for parity, and report compute unit deltas. It’s off by default (feature flag), includes a simple CLI command (mollusk matrix) with --variant name:b58,b58 syntax, and can emit Markdown/JSON reports with thresholds to flag regressions. This helps validate feature activations quickly without a validator and keeps default users unaffected.

tanmay4l:
fixed #160


TanmayDhobale:
@buffalojoec can u review this pr if possible

README.md
```diff
@@ -500,3 +500,27 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
 
 Fixtures can be loaded from files or decoded from raw blobs. These
 capabilities are provided by the respective fixture crates.
+
+## Feature Matrix (gated)
```
buffalojoec: You can just make a CLI section, since we need to add more CLI docs eventually.

```suggestion
## CLI

Mollusk CLI (WIP).

### Feature Matrix
```

README.md
```diff
@@ -500,3 +500,27 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
...
+
+The Feature Matrix lets you run fixtures under a baseline FeatureSet and automatically generated feature variants, then compare parity and compute units.
+
+- Gated behind the `feature-matrix` feature in `mollusk-svm` and `mollusk-svm-cli`.
+- Baseline FeatureSet is derived from each fixture’s `SlotContext`/`Context.feature_set` (no CLI slot override).
```
buffalojoec: I don't think this line makes a lot of sense. The fixtures provide the features that should be enabled, otherwise they're disabled. The matrix will toggle the targeted features on and off in separate runs.

README.md
```diff
@@ -500,3 +500,27 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
...
+CLI example:
+
+```bash
+mollusk matrix \
+  ./programs/token/src/elf/token.so \
+  ./harness/tests \
+  11111111111111111111111111111112 \
+  --feature 2yDdYzg56LgRBzX1UeMNcrsXphJ4yTZe4cCvEoGzbXDc \
+  --feature 5GDFSoKFJWWEkttfwTwWXfehKH7DGiL9v8bNChjJ5Q5g \
+  --matrix cartesian \
```
buffalojoec: I don't think this should be an argument, rather just the behavior itself.

cli/Cargo.toml
```diff
@@ -18,11 +18,20 @@ mollusk-svm-bencher = { workspace = true }
 mollusk-svm-fuzz-fixture = { workspace = true }
 mollusk-svm-fuzz-fixture-firedancer = { workspace = true }
 mollusk-svm = { workspace = true, features = ["fuzz", "fuzz-fd", "serde"] }
+bs58 = { workspace = true, optional = true }
```
buffalojoec: Can you please sort dependencies alphabetically?

cli/Cargo.toml
```diff
@@ -18,11 +18,20 @@ mollusk-svm-bencher = { workspace = true }
...
 solana-sdk-ids = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 
+[features]
+default = []
+feature-matrix = [
+    "bs58",
+    "mollusk-svm/feature-matrix",
+    "mollusk-svm/serde",
+]
```
buffalojoec: Any particular reason to feature-flag it? It wouldn't be bad to just cook it into the main CLI.

cli/src/main.rs
```diff
@@ -107,6 +109,33 @@ enum SubCommand {
         #[arg(short, long)]
         verbose: bool,
     },
+    #[cfg(feature = "feature-matrix")]
+    Matrix {
```
buffalojoec, line 116: Can you put doc comments on each field, just like the other CLI commands do? This lets clap put help documentation on each argument.

cli/src/main.rs
```diff
@@ -107,6 +109,33 @@ enum SubCommand {
...
+        #[arg(required = true)]
+        elf_path: String,
+        #[arg(required = true)]
+        fixtures_dir: String,
```
buffalojoec: Let's keep the field names and how they work consistent across CLI commands.

cli/src/main.rs
buffalojoec: Note you can basically pull most of the inspiration for this command's interface from `ExecuteFixture`.

cli/src/main.rs
```diff
@@ -107,6 +109,33 @@ enum SubCommand {
...
+        #[arg(value_parser = Pubkey::from_str)]
+        program_id: Pubkey,
+        #[arg(long)]
+        variant: Vec<String>,
+        #[arg(long)]
+        feature: Vec<String>,
```
buffalojoec: Why do we need both of these?

cli/src/main.rs
```diff
@@ -107,6 +109,33 @@ enum SubCommand {
...
+        #[arg(long, default_value = "cartesian")]
+        matrix: MatrixStrategy,
```
buffalojoec: We don't need this. It's a singleton anyway.

cli/src/main.rs
```diff
@@ -107,6 +109,33 @@ enum SubCommand {
...
+        #[arg(long)]
+        max_cu_delta_abs: Option<u64>,
+        #[arg(long)]
+        max_cu_delta_percent: Option<f32>,
+        #[arg(long)]
+        out_dir: Option<String>,
```
buffalojoec, line 140: Again, please be consistent with other CLI command fields.

harness/src/feature_matrix.rs
buffalojoec, line 1: Why does any of this need to go into the harness?

Why not just - totally within the CLI - create a `Mollusk` instance and mutate its `feature_set` directly?

cli/src/main.rs
```diff
@@ -227,6 +262,215 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+            let mut variants_list: Vec<FeatureVariant> = Vec::new();
+            for spec in &variant {
+                let (name, list) = match spec.split_once(':') {
+                    Some(v) => v,
+                    None => {
+                        eprintln!(
+                            "[matrix] ignoring malformed --variant (missing ':'): {}",
+                            spec
+                        );
```
buffalojoec: Is this logging pattern used anywhere else in the library?

cli/src/main.rs
```diff
@@ -227,6 +262,215 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+            let fm = FeatureMatrix::new(
+                mollusk,
+                BaselineConfig::Explicit(mollusk_svm::Mollusk::default().feature_set.clone()),
+            )
+            .thresholds(Thresholds {
+                max_cu_delta_abs,
+                max_cu_delta_percent,
+                require_result_parity: true,
+            });
```
buffalojoec: Why not just treat the fixture itself as the base? Whatever features are enabled in the test, that's the base case. Then, any toggling on/off that happens from the Cartesian product are treated as variants.

cli/src/main.rs
```diff
@@ -227,6 +262,215 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
+            let mut variants_list: Vec<FeatureVariant> = Vec::new();
+            for spec in &variant {
```
buffalojoec: Feels like we really don't need `--variant` and `--feature`, only one. I think it's easier to just provide a bunch of features you want to be variant over, ie. `--feature <>`.

README.md
tanmay4l: Added CLI section with Feature Matrix subsection

README.md
tanmay4l: okay ive clarified that fixtures define baseline, matrix toggles features

TanmayDhobale:
**@buffalojoec** Thanks for the feedback! I've addressed all the review comments:

 Added CLI section with Feature Matrix subsection  
 Clarified that fixtures define baseline, matrix toggles features  
 Cartesian generation is now the default behavior  
 Matrix command is always available  
 Only --feature multi-flag remains  
All matrix args now have comprehensive documentation  
 Using consistent `elf_path`, `fixture`, `program_id` across commands  
Replaced eprintln with println for consistency  
Cleaned up cli/Cargo.toml  

**Regarding the feature matrix architecture**: I kept the current implementation in the harness rather than moving it to the CLI. The 533-line module is working well, has comprehensive tests, and could be useful for other tools in the future. If you'd prefer it moved to the CLI, I can do that too , just let me know!

README.md
```diff
@@ -500,3 +500,30 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
...
+## CLI
+
+Mollusk CLI (WIP).
+
+### Feature Matrix
+
+The Feature Matrix lets you run fixtures under a baseline FeatureSet and automatically generated feature variants, then compare parity and compute units.
+
+- Baseline FeatureSet is derived from each fixture's `SlotContext`/`Context.feature_set`. The matrix toggles the targeted features on and off in separate runs.
+- Cartesian product generation is the default behavior for feature combinations.
```
buffalojoec: I still think this isn't very clear documentation about what the feature does. It should articulate what a matrix is in this context, how it works with the CLI tooling, and why you might want to use it.

For example: "...enables you to test how your program behaves when different Solana features are activated".

cli/Cargo.toml
```diff
@@ -9,20 +9,24 @@ edition = { workspace = true }
...
+serde_json = { workspace = true }
+serde_yaml = { workspace = true }
 solana-logger = { workspace = true }
 solana-pubkey = { workspace = true }
 solana-sdk-ids = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 
+[features]
+default = []
+
```
buffalojoec: We don't need this, since there are no features!

cli/src/main.rs
buffalojoec: Looks like we're still missing some doc comments.

cli/src/main.rs
buffalojoec: Looks like this hasn't been resolved yet, either?

cli/src/main.rs
```diff
@@ -227,6 +259,178 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                    }
+                }
+            }
+
+            let fixtures = search_paths(&fixture, "fix")?;
+            if matches!(proto, ProtoLayout::Firedancer) {
+                println!(
+                    "[matrix] Firedancer fixtures are not supported by the matrix runner yet. Use \
+                     --proto mollusk."
+                );
```
buffalojoec: This logging pattern still does not match what the rest of the library does.

cli/src/main.rs
```diff
@@ -227,6 +259,178 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                std::process::exit(1);
```
buffalojoec: This should just return an `Err` and let the `main` process handle the exit.

cli/src/main.rs
```diff
@@ -227,6 +259,178 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            }
+
+            use std::collections::HashMap;
```
buffalojoec: Why the in-line `use` statement here?

cli/src/main.rs
```diff
@@ -227,6 +259,178 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                {
+                    let features = base_fs.clone();
+                    let mut m = fm.build_mollusk_for_featureset(&features);
+                    add_elf_to_mollusk(&mut m, &elf_path, &program_id);
+                    let res = m.process_fixture(&fixture);
+                    let entry = aggregates.entry("baseline".to_string()).or_default();
+                    let mut total = entry.clone();
+                    total.absorb(res);
+                    *entry = total;
+                }
```
buffalojoec: I don't think you need all of these clones.

cli/src/main.rs
```diff
@@ -227,6 +259,178 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            }
+
+            let mut runs: Vec<
+                mollusk_svm::feature_matrix::VariantRun<mollusk_svm::result::InstructionResult>,
+            > = Vec::new();
+            let meta_base = fm.resolve_baseline_featureset();
+            runs.push(mollusk_svm::feature_matrix::VariantRun {
+                name: "baseline".to_string(),
+                features: meta_base.clone(),
```
buffalojoec: I think this should be `fs_base`, no?

cli/src/main.rs
```diff
@@ -227,6 +259,178 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                output: aggregates.remove("baseline").unwrap_or_default(),
+            });
+            for v in variants_list {
+                runs.push(mollusk_svm::feature_matrix::VariantRun {
+                    name: v.name.clone(),
+                    features: meta_base.clone(),
```
buffalojoec: This one, too?

harness/Cargo.toml
```diff
@@ -25,7 +25,6 @@ fuzz-fd = [
 ]
 invocation-inspect-callback = []
 serde = [
-    "dep:serde",
```
buffalojoec, line 28: Let's make sure we get the optional serde back.

cli/src/main.rs
```diff
@@ -227,6 +259,178 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            let mut runs: Vec<
+                mollusk_svm::feature_matrix::VariantRun<mollusk_svm::result::InstructionResult>,
+            > = Vec::new();
+            let meta_base = fm.resolve_baseline_featureset();
+            runs.push(mollusk_svm::feature_matrix::VariantRun {
+                name: "baseline".to_string(),
+                features: meta_base.clone(),
+                output: aggregates.remove("baseline").unwrap_or_default(),
+            });
+            for v in variants_list {
```
buffalojoec: We should probably have an upper limit on the number of variants one can create. It might generate an insane amount of variants and run for a long time.

harness/src/feature_matrix.rs
```diff
@@ -0,0 +1,544 @@
...
+            None
+        };
+
+        if let Some(cfg) = &self.report {
+            if let Some(dir) = &cfg.out_dir {
+                let _ = std::fs::create_dir_all(dir);
+                if cfg.markdown {
+                    if let Some(md) = &markdown {
+                        let path = dir.join("feature_matrix.md");
+                        let _ = std::fs::write(path, md);
```
buffalojoec: This means we'll have no way to know what happened if the file failed to write.

tanmay4l: ig its better to moved the matrix into the CLI and now ill report writes return an error on dir creation failure and log file-write failures, so nothing fails silently

cli/src/main.rs
tanmay4l: should we put hard cap (4096 variants); and if exceeded,  command errors with a clear message

harness/Cargo.toml
tanmay4l: sure Restored the serde

cli/src/main.rs
tanmay4l: yep that was wrong UX-wise—the true baseline is the fixture’s FeatureSet. Using fs_base

cli/src/main.rs
tanmay4l: thinking to replaced cloning with direct use of fs_base and in-place entry.absorb(res), removing unnecessary FeatureSet and aggregate clones

cli/src/main.rs
tanmay4l: moved the inline use std::collections::HashMap; to the top imports for consistency

cli/src/main.rs
tanmay4l: Updated to return an Err from the command arm;

cli/src/main.rs
tanmay4l: removed the println and now return an Err from the command;

TanmayDhobale:
@buffalojoec  moved the matrix engine to the CLI cleaned dead code, tried clarified README, and bubbled report I/O as errors; happy to follow up with a separate PR for report polish

harness/Cargo.toml
```diff
@@ -35,6 +35,7 @@ agave-precompiles = { workspace = true }
 agave-syscalls = { workspace = true }
 bincode = { workspace = true }
 serde = { workspace = true, features = ["derive"], optional = true }
+serde_json = { workspace = true, optional = true }
```
buffalojoec: I don't think we need this, right? The harness should be untouched by this PR.

cli/src/main.rs
```diff
@@ -107,6 +108,43 @@ enum SubCommand {
...
+        out_dir: Option<String>,
+        #[arg(long)]
+        markdown: bool,
+
+        /// Emit a JSON report.
+        #[arg(long)]
+        json: bool,
+
+        #[arg(long, default_value = "mollusk")]
+        proto: ProtoLayout,
```
buffalojoec, line 147:
```suggestion
        /// Protobuf layout to use when executing the fixture.
        #[arg(long, default_value = "mollusk")]
        proto: ProtoLayout,
```

cli/src/main.rs
```diff
@@ -147,6 +185,8 @@ fn add_elf_to_mollusk(mollusk: &mut Mollusk, elf_path: &str, program_id: &Pubkey
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Initialize logging for consistent output from log::* macros.
+    solana_logger::setup_with("");
```
buffalojoec: Why is this necessary? Each `Mollusk` instance will initialize logging.

cli/src/main.rs
```diff
@@ -227,6 +267,176 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            program_id,
+            feature,
+            max_cu_delta_abs,
+            max_cu_delta_percent,
+            out_dir,
+            markdown,
+            json,
+            proto,
+        } => {
+            use {crate::matrix, bs58};
```
buffalojoec: Please put this up top.

cli/src/main.rs
```diff
@@ -107,6 +108,43 @@ enum SubCommand {
...
+
+        /// Output directory for generated reports.
+        #[arg(long)]
+        out_dir: Option<String>,
+        #[arg(long)]
+        markdown: bool,
+
+        /// Emit a JSON report.
+        #[arg(long)]
+        json: bool,
```
buffalojoec: What do you think about changing the format args to be something like this?

```
<cmd> --output json --output markdown
```

Then you can just collect all output formats, and default to just markdown if there are no `--output` flags provided. What do you think?

cli/src/main.rs
```diff
@@ -227,6 +267,176 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            });
+
+            let mut variants_list: Vec<matrix::FeatureVariant> = Vec::new();
+            if !feature.is_empty() {
+                let feature_ids: Vec<solana_pubkey::Pubkey> = feature
+                    .iter()
+                    .filter_map(|s| bs58::decode(s).into_vec().ok())
+                    .filter(|v| v.len() == 32)
+                    .map(|v| solana_pubkey::Pubkey::new_from_array(v.try_into().unwrap()))
+                    .collect();
```
buffalojoec: I have an idea of how we could optimize this workflow here.

1. Within the `main` module (here, around where my comment is), we could simply collect all of the provided features as `Pubkey`, and throw an error if none were provided.
2. Then, we can simply pass that list of pubkeys along to a helper, which will create a list of variants to apply to the `Mollusk` instance.

A variant could look something like this:

```rust
struct Variant {
    pub apply: Fn(&mut FeatureSet),
    pub reset: Fn(&mut FeatureSet),
}

let mut mollusk = Mollusk::new(..);
let baseline_result = mollusk.process_instruction(..);

let mut variant_results = Vec::new();
for variant in variants {
    variant.apply(&mut mollusk.feature_set);
    let result = mollusk.process_instruction(..);
    variant_results.push(result);
    variant.reset(&mut mollusk.feature_set);
}
```

buffalojoec: Furthermore, it might not be a bad idea to limit this CLI command to [runtime features](https://github.com/anza-xyz/agave/blob/1cc30112719408a7b3b688ead983eecd377824dc/feature-set/src/lib.rs#L103). That way, we can actually use `Fn(&mut SVMFeatureSet)`, which is all `bool` values, so it's a smaller memory footprint to manipulate.

cli/src/main.rs
buffalojoec: Did you see the `verbose` option on the other commands? That's a good pattern to follow if you want logging.

cli/src/main.rs
buffalojoec: See the example I wrote out. I think that addresses the in-place mutation idea.

cli/src/main.rs
buffalojoec: Yeah sure, that sounds good.

cli/src/main.rs
tanmay4l: Added the proto doc comments as suggested.
Dropped the global logging init; Mollusk initializes logging per instance.

cli/src/main.rs
tanmay4l: done

cli/src/main.rs
tanmay4l: yea works looks good imo , Moved the inline import to the top as requested; thanks for the catch.

cli/src/main.rs
tanmay4l: yea that’s a great call. I’ve updated the matrix runner to:

- Collect all provided features as Pubkeys up front (error if none valid)
- Generate a set of Variants with apply/reset closures
- Run a single Mollusk per fixture: apply → execute → reset
- Keep a hard cap of 4096 variants

right now I’m toggling via FeatureSet.active_mut().insert/remove. If you’d prefer we constrain to runtime-only toggles and flip SVMFeatureSet boolean flags instead, I can switch that over—point me at the exact API you want used and I’ll wire it up.

PR #171, opened 2025-11-03
New crate: `mollusk-svm-on-demand`

L0STE:
This crate simplifies testing by fetching and storing mainnet accounts directly in the `AccountStore`.

## Key Features

The new `RpcAccountStore` allows you to:

- **Pass mock accounts** using the `with_accounts` method
- **Fetch mainnet accounts from instructions** using the `from_instruction` method to automatically dump all accounts referenced in an instruction
- **Handle missing accounts gracefully** with the `allow_missing_accounts` method, which stores them with default values
- **Add CPI programs** using the `add_programs` method, which fetches program data from the account store and loads it via `mollusk.add_program_with_elf_and_loader` (supports both loader V2 and V3)
- **Sync slots with mainnet** using `with_synced_slot` for oracles that require the current mainnet slot to function correctly

## New Harness

Added a new test harness variant with `*_with_context` methods that implement all the above helpers. These methods support both strict and non-strict modes depending on how they're called, while maintaining the same interface as the original harness.

### DIFF
--- a/on-demand/src/account_store.rs
+++ b/on-demand/src/account_store.rs
@@ -358,12 +358,4 @@ impl RpcAccountStore {
         mollusk.warp_to_slot(slot);
         Ok(self)
     }
-
-    /// Deprecated: Use `with_synced_slot` instead.
-    ///
-    /// This method will be removed in a future version.
-    #[deprecated(since = "0.0.1", note = "Use `with_synced_slot` instead")]
-    pub async fn sync_slot(self, mollusk: &mut Mollusk) -> Result<Self, RpcError> {
-        self.with_synced_slot(mollusk).await
-    }
 }


on-demand/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "mollusk-svm-on-demand"
```
buffalojoec, line 2: I think we need to bikeshed the name. 😅

This library tends to use more descriptive (sometimes verbose) naming conventions, so it's clear what a given crate does. "On demand" is pretty vague in my opinion.

What about something like one of these?
* `mollusk-account-store-on-demand`
* `mollusk-account-store-rpc-on-demand`
* `mollusk-account-store-rpc`

on-demand/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+description = "Automatically fetch and use mainnet accounts with the Mollusk SVM harness."
```
buffalojoec, line 3: Is it only mainnet-beta?

on-demand/src/account_store.rs
```diff
@@ -0,0 +1,361 @@
+//! RPC account store for fetching accounts from Solana RPC endpoints.
+//!
+//! This module provides the `RpcAccountStore` type for automatically fetching
+//! accounts from mainnet and managing them for use with Mollusk testing.
```
buffalojoec, line 4: It doesn't appear the tool is limited to mainnet. Why should it be, anyway?

on-demand/src/account_store.rs
```diff
@@ -0,0 +1,361 @@
...
+        mut self,
+        instructions: &[Instruction],
+    ) -> Result<Self, RpcError> {
+        let pubkeys: HashSet<Pubkey> = instructions
+            .iter()
+            .flat_map(|ix| ix.accounts.iter().map(|m| m.pubkey))
+            .collect();
+
+        self.fetch_accounts(&pubkeys.into_iter().collect::<Vec<_>>())
+            .await?;
```
buffalojoec, line 171: Since `fetch_accounts` is internal, you could just make it accept a pubkey iterator (`impl Iterator<Item = Pubkey>`). You're also deduping in the `fetch_accounts` as well, so no need for the hash map above this. Just pipe the mapped iterator over the instruction accounts directly into `fetch_accounts`!

on-demand/src/account_store.rs
```diff
@@ -0,0 +1,361 @@
...
+            .iter()
+            .filter(|pubkey| !self.cache.contains_key(pubkey))
+            .copied()
+            .collect();
+
+        if missing_pubkeys.is_empty() {
+            return Ok(());
+        }
+
+        let accounts = self.client.get_multiple_accounts(&missing_pubkeys).await?;
```
buffalojoec, line 199: Eh, I suppose you maybe do have to do one allocation, but perhaps you can filter by `cache.contains_key` and collect that into a vec, then pass to the RPC call?

on-demand/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+        .with_accounts(accounts)
+        .from_instruction(instruction)
+        .await?
+        .add_programs(&mut mollusk)
+        .await?
+        .cache;
+
+    let context = mollusk.with_context(cache);
+    Ok(context.process_instruction(instruction))
+}
```
buffalojoec, line 83: Why not just let developers create an `RpcAccountStore` and then do this?

```rust
let store = RpcAccountStore::new(rpc_url)
        .allow_missing_accounts()
        .skip_program_validation();

let mollusk = Mollusk::default().with_context(&store);

mollusk.process_instruction(
    &instruction,
    &accounts,
);
```

You just need to implement `AccountStore` for `RpcAccountStore`.
https://github.com/anza-xyz/mollusk/blob/fc7724585678c5f14cdb7b3ded15fe3533a968b6/harness/src/account_store.rs#L7

Is it because your fetcher is `async`? Can you use the blocking RPC client instead?

on-demand/src/account_store.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        Ok(())
+    }
+
+    /// Add programs to the Mollusk environment.
+    ///
+    /// This function fetches the program data accounts for all programs that are
+    /// stored in the cache and adds them to the Mollusk environment.
+    ///
+    /// Note: This is needed because mollusk-svm doesn't load the programs for CPIs directly from the accounts.
```
buffalojoec, line 227: Hm, maybe it should (?)

L0STE:
> My biggest concern is that it reimplements a lot of the harness's API.

Agree, I don't love that as well. 

Tbh all current methods works without having to reimplement the harness, I added that after just to make it easier to use for people that don't really know how to do so. But probably it's just more confusing and just midcurved the entire implementation.

I'm down to nuke the reimplementation of the harness

on-demand/src/lib.rs
L0STE: Yes I maybe should have made it more clear but this is currently possible. 

You just need to add an `.await()` at the end of the store implementation and you can then blast it in the normal harness

on-demand/Cargo.toml
L0STE: Good catch, it's not. Fixing

PR #172, opened 2025-11-05
chore: bump solana deps

sonicfromnewyoke:
### Problem

Mollusk currently uses outdated solana deps that leads to importing of more libs

### Summary of Changes

bump solana version

cc @buffalojoec


buffalojoec (approved):
Lgtm thanks!

PR #178, opened 2025-11-21
Register tracing export

procdump:
**Problem**

With solana-sbpf 0.13.1 it's possible to collect register tracing data in a consistent way. Yet a testing harness like mollusk does not provide a default mechanism to collect this data in a useful format for later ingestion.
The idea of this PR is to provide such mechanism in a way consumer tools could take advantage of - particularly willing to analyze the tracing data and generate statistics on top of it.

**Solution**

The PR takes advantage of the `invocation-inspect-callback` feature where a callback could be called prior to or after the processed instruction. A default post instruction callback for storing the register tracing data in files is provided and it basically persists the register sets, sbpf instructions and an identification sha256 hash of the executable that was actually used for the generation of this data. The motivation behind providing the sha256 identification is that files could grow in number and it's necessary for the consumer to be able to distinguish which shared object to use when analyzing the tracing data in a deterministic way. Also in order to let consumers instantly take advantage of this feature it's written in a way that register tracing gets enabled as long as the `SBF_TRACE_DIR` is provided denoting the directory in which the tracing data will be persisted. The motivation behind this approach is that it will let users instantly get out the tracing data of their tests which can eventually be passed to statistics generators such as code coverage tools, disassemblers, etc.

**Review**

Before getting further I'd be happy to get some feedback if you find this approach feasible or have a better idea how it may be done.

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -714,7 +714,7 @@ impl Mollusk {
         enable_register_tracing: bool,
     ) -> Self {
         let mut mollusk = Self::new_inner(enable_register_tracing);
-        mollusk.add_program(program_id, program_name, &DEFAULT_LOADER_KEY);
+        mollusk.add_program(program_id, program_name);
         mollusk
     }
 


harness/src/file.rs
```diff
@@ -27,7 +27,7 @@ use {
     },
 };
 
-fn default_shared_object_dirs() -> Vec<PathBuf> {
+pub fn default_shared_object_dirs() -> Vec<PathBuf> {
```
buffalojoec: Does keeping this crate-private work?

```suggestion
pub(crate) fn default_shared_object_dirs() -> Vec<PathBuf> {
```

harness/src/lib.rs
```diff
@@ -509,6 +511,8 @@ pub struct Mollusk {
     /// programs comes from the sysvars.
     #[cfg(feature = "fuzz-fd")]
     pub slot: u64,
+
+    pub enable_register_tracing: bool,
```
buffalojoec: If this is always going to be `false` when the feature is disabled, I think it should be private, right?

harness/src/lib.rs
```diff
@@ -658,6 +683,26 @@ impl Mollusk {
...
+            &mollusk.compute_budget,
+            mollusk.enable_register_tracing,
+        );
+        mollusk.program_cache = program_cache;
+        mollusk.invocation_inspect_callback =
+            Box::new(register_tracing::DefaultRegisterTracingCallback {});
+        mollusk.add_program(program_id, program_name, &DEFAULT_LOADER_KEY);
+        mollusk
+    }
+
```
buffalojoec: Do you need this new constructor? The existing `Mollusk::new` will call into `Self::default`, and the logic gate where you check for the `SBF_TRACE_DIR` environment variable will be evaluated. So people can just use `Mollusk::new`!

harness/src/register_tracing.rs
```diff
@@ -0,0 +1,132 @@
...
+    instruction_context: InstructionContext,
+    executable: &Executable,
+    register_trace: RegisterTrace,
+) -> Result<(), Box<dyn std::error::Error>> {
+    if register_trace.is_empty() {
+        // Can't do much with an empty trace.
+        return Ok(());
+    }
+
+    if let Ok(sbf_trace_dir) = &std::env::var("SBF_TRACE_DIR") {
```
buffalojoec: Since register tracing is only ever enabled when `SBPF_TRACE_DIR` is set, you might as well store `sbf_trace_dir` on the `DefaultRegisterTracingCallback` struct itself, since it's only instantiated when the tracing is enabled. Then you can eliminate the need for this big `if let Ok(..)` block.

harness/src/file.rs
procdump: Right, fixing it.

harness/src/lib.rs
procdump: Making it private prevents such manual instantiations where we get a compiler complaint that the enable_register_tracing is private: https://github.com/anza-xyz/mollusk/blob/fc7724585678c5f14cdb7b3ded15fe3533a968b6/harness/tests/fd_test_vectors.rs#L69

buffalojoec: Ah crud. What about just using the environment variable everywhere for the logic gate? The problem is, if someone flips this to `true` on Mollusk directly, but doesn't provide the `SBF_TRACE_DIR`, it's going to disable tracing anyway. This would be confusing for people.

Alternatively, we could choose a default dir under `target` if we really want this to be public. However, doing something like this:

```rust
let mut mollusk = Mollusk::default();
mollusk.enable_register_tracing = true;
...
```

...would still not add the callback and not activate tracing, as per the current implementation.

procdump: > Ah crud. What about just using the environment variable everywhere for the logic gate? The problem is, if someone flips this to `true` on Mollusk directly, but doesn't provide the `SBF_TRACE_DIR`, it's going to disable tracing anyway. This would be confusing for people.

You're right - it would be confusing. There's another case that also concerns me -  it's related to the environment variable only approach. As I'm going to write some tests for this feature it's that I would need to set the envvar from within the test itself. Setting envvars is unsafe as tests are run in the same process but in different threads. Hence having multiple tests playing with the SBF_TRACE_DIR var is UB. This is the reason I added with_register_tracing in the first place - just having a constructor that has the ability to instantiate mollusk with register tracing but without the need to play with the envvar.

> 
> Alternatively, we could choose a default dir under `target` if we really want this to be public. However, doing something like this:

I think it's probably a good idea to have a default path.

> 
> ```rust
> let mut mollusk = Mollusk::default();
> mollusk.enable_register_tracing = true;
> ...
> ```
> 
> ...would still not add the callback and not activate tracing, as per the current implementation.

I think I came up with some sort of a possible solution - I can bury the enable_register_tracing bool in the callback structure - thus nobody would be able to set it without actually providing a handler - I'll try to rework it to see the final outcome and will show it here.

buffalojoec: > Setting envvars is unsafe as tests are run in the same process but in different threads. Hence having multiple tests playing with the SBF_TRACE_DIR var is UB.

Why wouldn't the `SBF_TRACE_DIR` variable just be readonly across all threads?

buffalojoec: > I think I came up with some sort of a possible solution - I can bury the enable_register_tracing bool in the callback structure - thus nobody would be able to set it without actually providing a handler - I'll try to rework it to see the final outcome and will show it here.

Cool! Ping me when you throw it up.

procdump: I've pushed the changes. I've also added one test implementing a custom register tracing callback counting the number of JMP class instructions executed. Tests help me reason about the API and its consumption later. On one side it's now that you can tweak existing codebases with the `SBF_TRACE_DIR` and on the other side it's possible to enable register tracing without the need to tweak the same environment variable at all - hence the with_register_tracing constructor.

What I don't like up to now is that you can instantiate Mollusk::default() and change the callback expecting this will enable register tracing. Unfortunately it won't because it's the ProgramCache that also needs to have register tracing tweaked. Yet doesn't such post-fact changes also not work for the rest of the Mollusk fields where there's a cross dependency between some of them?

procdump: > > Setting envvars is unsafe as tests are run in the same process but in different threads. Hence having multiple tests playing with the SBF_TRACE_DIR var is UB.
> 
> Why wouldn't the `SBF_TRACE_DIR` variable just be readonly across all threads?

It could be but if we need to enable tracing for one test and skip it for another the envvar is not very convenient IMO.

harness/src/lib.rs
procdump: > Do you need this new constructor? The existing `Mollusk::new` will call into `Self::default`, and the logic gate where you check for the `SBF_TRACE_DIR` environment variable will be evaluated. So people can just use `Mollusk::new`!

I'd be glad to have your review up to now and if you prefer I'll rework it to have everything gated behind the environment variable - I have in mind something like this:
```patch
diff --git a/harness/src/lib.rs b/harness/src/lib.rs
index 815ec07..b89830f 100644
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -689,32 +689,6 @@ impl Mollusk {
         mollusk
     }
 
-    /// Create a new Mollusk instance just like the `new` method but
-    /// with register tracing enabled using a default callback.
-    ///
-    /// If `SBF_TRACE_DIR` is set it will override the passed `sbf_trace_dir`.
-    #[cfg(feature = "register-tracing")]
-    pub fn with_register_tracing(
-        program_id: &Pubkey,
-        program_name: &str,
-        sbf_trace_dir: &str,
-    ) -> Self {
-        let mut mollusk = Mollusk::default();
-        mollusk.invocation_inspect_callback =
-            Box::new(register_tracing::DefaultRegisterTracingCallback {
-                sbf_trace_dir: std::env::var("SBF_TRACE_DIR").unwrap_or(sbf_trace_dir.into()),
-            });
-        mollusk.program_cache = ProgramCache::new(
-            &mollusk.feature_set,
-            &mollusk.compute_budget,
-            mollusk
-                .invocation_inspect_callback
-                .is_register_tracing_callback(),
-        );
-        mollusk.add_program(program_id, program_name, &DEFAULT_LOADER_KEY);
-        mollusk
-    }
-
     /// Add a program to the test environment.
     ///
     /// If you intend to CPI to a program, this is likely what you want to use.
diff --git a/harness/src/register_tracing.rs b/harness/src/register_tracing.rs
index ddb3cc0..7c18b1e 100644
--- a/harness/src/register_tracing.rs
+++ b/harness/src/register_tracing.rs
@@ -85,7 +85,7 @@ impl InvocationInspectCallback for DefaultRegisterTracingCallback {
 
     // This callback is specifically implemented to handle register tracing.
     fn is_register_tracing_callback(&self) -> bool {
-        true
+        std::env::var("SBF_TRACE_DIR").is_ok()
     }
 }
 
diff --git a/harness/tests/register_tracing.rs b/harness/tests/register_tracing.rs
index ba9e85c..0c30d15 100644
--- a/harness/tests/register_tracing.rs
+++ b/harness/tests/register_tracing.rs
@@ -87,11 +87,11 @@ fn test_custom_register_tracing_callback() {
     }
 
     std::env::set_var("SBF_OUT_DIR", "../target/deploy");
+    std::env::set_var("SBF_TRACE_DIR", "/tmp/sbf_trace_dir");
 
     let program_id = Pubkey::new_unique();
     let payer_pk = Pubkey::new_unique();
-    let mut mollusk =
-        Mollusk::with_register_tracing(&program_id, "test_program_primary", "sbf_trace_dir");
+    let mut mollusk = Mollusk::new(&program_id, "test_program_primary");
 
     // Phase 1 - basic register tracing test.
 
```

harness/tests/register_tracing.rs
```diff
@@ -0,0 +1,177 @@
...
+    // Phase 2 - check we can stop register tracing for this instance of Mollusk.
+    {
+        // Clear the tracing data collected so far.
+        {
+            let mut td = tracing_data.borrow_mut();
+            td.clear();
+        }
+
+        mollusk.enable_register_tracing = false;
+
```
procdump: > Added your enable_register_tracing field back to Mollusk, for disabling the trace during instruction invocation.

There's a subtle discrepancy here. I've reworked the test so that I can show you what I mean and eventually you can debug it yourself.
Let's say you want to set `enable_register_tracing` to `false` and you'd expect that this will disable the trace during instruction invocation because in `process_instruction_inner` the `false` value is passed through `InvokeContext`:

```rust
            let _enable_register_tracing = false;
            #[cfg(feature = "register-tracing")]
            let _enable_register_tracing = self.enable_register_tracing;

            let program_runtime_environments: ProgramRuntimeEnvironments =
                ProgramRuntimeEnvironments {
                    program_runtime_v1: Arc::new(
                        create_program_runtime_environment_v1(
                            &runtime_features,
                            &execution_budget,
                            /* reject_deployment_of_broken_elfs */ false,
                            /* debugging_features */ _enable_register_tracing,
                        )
                        .unwrap(),
                    ),
                    program_runtime_v2: Arc::new(create_program_runtime_environment_v2(
                        &execution_budget,
                        /* debugging_features */ _enable_register_tracing,
                    )),
                };

            let mut invoke_context = InvokeContext::new(
                &mut transaction_context,
                &mut program_cache,
                EnvironmentConfig::new(
                    Hash::default(),
                    /* blockhash_lamports_per_signature */ 5000, // The default value
                    &callback,
                    &runtime_features,
                    &program_runtime_environments,
                    &program_runtime_environments,
                    &sysvar_cache,
                ),
                self.logger.clone(),
                self.compute_budget.to_budget(),
                self.compute_budget.to_cost(),
            );
```
However register tracing still remains enabled surprisingly. The reason for this goes deep into the runtime - particularly in the execution of the instruction within the VM (let's look in the case of being into interpreted mode):
https://github.com/anza-xyz/sbpf/blob/c0c1197edbd1379a2f1ae86140fec30ec46b08e9/src/interpreter.rs#L176
https://github.com/anza-xyz/sbpf/blob/c0c1197edbd1379a2f1ae86140fec30ec46b08e9/src/interpreter.rs#L190
Basically it reads `enable_register_tracing` from the `Executable` instead of `InvokeContext's` `ProgramRuntimeEnvironments`.

Due to the circumstances upper the register tracing unit test fails now. Isn't it more appropriate to read the flag from `InvokeContext's EnvironmentConfig's program_runtime_v2's Config`?
With such a change we'd be able to switch on/off tracing as you'd suggested.
Otherwise it's confusing for users that if they tweak the flag it won't influence the tracing.

NB: I've pushed the amended test on purpose so that you see that it fails.

> Considering what I said in bullet 1, we actually could just remove it from add_program completely. What do you think?

Doesn't this suggestion depend on fixing the issue explained upper?

buffalojoec: Ahh - thanks for digging into that. This is tricky.

> Isn't it more appropriate to read the flag from `InvokeContext's EnvironmentConfig's program_runtime_v2's Config`?

You mean on the `sbpf` side? That does make sense to me! It seems like we'd have to decouple which `Config` is used for interpreter vs. JIT to get that to work. If you think you can fix it without too much churn I'm happy to help/review!

Even if we decide to fix that, I think we should consider what's a good way to get this feature merged in the meantime. The way I see it, we have three options:

1. Couple register tracing to the `register-tracing` feature flag and disallow toggling on and off.
2. Add a separate constructor behind `register-tracing` and use a construction-time only config. No toggling on an existing `Mollusk` instance.
3. Allow developers to _think_ they're disabling the tracing via some config - while `register-tracing` is enabled - but it really only changes the callback. It's a bit deceptive.

Ultimately I decided to give Number 2 a shot. What do you think? Can your workflow manage with the toggling as a constructor parameter, or do you need on-the-fly?

procdump: > You mean on the sbpf side? That does make sense to me! It seems like we'd have to decouple which Config is used for interpreter vs. JIT to get that to work. If you think you can fix it without too much churn I'm happy to help/review!

> Even if we decide to fix that, I think we should consider what's a good way to get this feature merged in the meantime. The way I see it, we have three options:

As you say it's probably best to focus on a good way to get the feature merged. Yet I'll try to check the possible approaches to having it decoupled.

> Ultimately I decided to give Number 2 a shot. What do you think? Can your workflow manage with the toggling as a constructor parameter, or do you need on-the-fly?

LGTM. I've only made the following changes:

- I've reintroduced the logic that if `SBF_TRACE_DIR` is set this will signal Mollusk to turn on tracing upon instantiation. I honestly believe that it's best to have it this way as this will virtually allow users to get traces without actually doing any changes to their code
- Fixed a comment which I believe you meant at first - s/compiled/loaded/
- A register tracing section in the README. Please have a look at it.

README.md
```diff
@@ -23,13 +23,15 @@ budget, feature set, or sysvars. These configurations are stored directly
...
-  * [Loading and Executing Fixtures](#loading-and-executing-fixtures)
+- [Mollusk](#mollusk)
+  - [Single Instructions](#single-instructions)
+  - [Instruction Chains](#instruction-chains)
+  - [Stateful Testing with MolluskContext](#stateful-testing-with-molluskcontext)
+  - [Benchmarking Compute Units](#benchmarking-compute-units)
+  - [Fixtures](#fixtures)
+    - [Generating Fixtures from Mollusk Tests](#generating-fixtures-from-mollusk-tests)
+    - [Loading and Executing Fixtures](#loading-and-executing-fixtures)
+  - [Register tracing](#register-tracing)
```
buffalojoec: Did you mean to change this list? I think we could just add your bullet at the end of it.

README.md
```diff
@@ -497,3 +499,54 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
...
+more fine-grained control over when register tracing data should be
+collected - for example, only for a specific instruction. Such control could
+be achieved by resetting the invocation callback to
+`EmptyInvocationInspectCallback` and later by restoring it to
+`DefaultRegisterTracingCallback`.
+
+```rust
+use mollusk_svm::{register_tracing, EmptyInvocationInspectCallback, Mollusk};
+
+assert!(std::env::var("SBF_TRACE_DIR").is_ok() == true);
```
buffalojoec:
```suggestion
assert!(std::env::var("SBF_TRACE_DIR").is_ok());
```

README.md
```diff
@@ -497,3 +499,54 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
...
+
+/* Mollusk setup .. */
+/* Load programs .. */
+
+/* .. */
+/* Have tracing here. */
+/* .. */
+
+/* Skip tracing from here. */
```
buffalojoec:
```suggestion
/* Disable trace collection here. */
```

README.md
```diff
@@ -497,3 +499,54 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
...
+mollusk.invocation_inspect_callback = Box::new(EmptyInvocationInspectCallback {});
+/* .. */
+
+/* Have tracing back. */
```
buffalojoec:
```suggestion
/* Re-enable trace collection. */
```

README.md
procdump: It seems the linter is the culprit - I'll bring it back in the former state with just the register tracing bullet added.

procdump:
All suggestions fixed and rebased - have a look now.

PR #180, opened 2025-11-27
fix: deterministic account compilation with bounds checking

### LINKED ISSUE #179: `compile_instruction_without_data` is non-deterministic and truncates
Invoking `compile_instruction_without_data` twice can result in different indices for the same account. If an index exceeds u8::MAX, it is truncated. 
https://github.com/anza-xyz/mollusk/blob/fc7724585678c5f14cdb7b3ded15fe3533a968b6/keys/src/accounts.rs#L18-L30
By comparison the solana-sdk uses a stable ordering and errors on truncation.
https://github.com/anza-xyz/solana-sdk/blob/a5ecbeb813096b11c63cca476acf9b4d49e27475/message/src/account_keys.rs#L109-L116

AvhiMaz:
Replace HashMap with BTreeMap for deterministic ordering and add explicit
overflow checks that return errors instead of silently truncating. Update
compile_instruction_without_data to return Result with error context.

Fixes #179

### DIFF
--- a/keys/src/accounts.rs
+++ b/keys/src/accounts.rs
@@ -15,34 +15,33 @@ pub struct CompiledInstructionWithoutData {
     pub accounts: Vec<u8>,
 }
 
-pub fn compile_instruction_without_data<'a>(
+pub fn compile_instruction_without_data(
     key_map: &KeyMap,
-    instruction: &'a Instruction,
-) -> Result<CompiledInstructionWithoutData, MolluskError<'a>> {
+    instruction: &Instruction,
+) -> CompiledInstructionWithoutData {
     let program_id_index = key_map
         .position(&instruction.program_id)
-        .ok_or(MolluskError::ProgramIdNotMapped(&instruction.program_id))
-        .and_then(|index| {
-            u8::try_from(index).map_err(|_| MolluskError::AccountIndexOverflow(index))
-        })?;
+        .or_panic_with(MolluskError::ProgramIdNotMapped(&instruction.program_id));
 
-    let accounts: Result<Vec<u8>, MolluskError<'a>> = instruction
+    let program_id_index = u8::try_from(program_id_index)
+        .or_panic_with(MolluskError::AccountIndexOverflow(program_id_index));
+
+    let accounts: Vec<u8> = instruction
         .accounts
         .iter()
         .map(|account_meta| {
-            key_map
+            let index = key_map
                 .position(&account_meta.pubkey)
-                .ok_or(MolluskError::AccountMissing(&account_meta.pubkey))
-                .and_then(|index| {
-                    u8::try_from(index).map_err(|_| MolluskError::AccountIndexOverflow(index))
-                })
+                .or_panic_with(MolluskError::AccountMissing(&account_meta.pubkey));
+
+            u8::try_from(index).or_panic_with(MolluskError::AccountIndexOverflow(index))
         })
         .collect();
 
-    Ok(CompiledInstructionWithoutData {
+    CompiledInstructionWithoutData {
         program_id_index,
-        accounts: accounts?,
-    })
+        accounts,
+    }
 }
 
 pub fn compile_instruction_accounts(
@@ -159,7 +158,7 @@ mod tests {
         let instruction = test_instruction(program_id, &[account1, account2]);
         let key_map = KeyMap::compile_from_instruction(&instruction);
 
-        let compiled = compile_instruction_without_data(&key_map, &instruction).unwrap();
+        let compiled = compile_instruction_without_data(&key_map, &instruction);
 
         assert_eq!(
             compiled.program_id_index,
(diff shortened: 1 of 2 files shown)


keys/src/accounts.rs
```diff
@@ -18,15 +18,39 @@ pub struct CompiledInstructionWithoutData {
...
-            .collect(),
+) -> Result<CompiledInstructionWithoutData, MolluskError<'static>> {
+    let program_id_index = key_map
+        .position(&instruction.program_id)
+        // Note: We use AccountIndexOverflow as the error type even for missing keys
+        // because MolluskError<'static> doesn't allow borrowed references. If the key
+        // is missing, returning an error (rather than panicking) allows the caller
+        // to handle it. The value passed to AccountIndexOverflow should be interpreted
+        // in context: if it equals the key_map length, the key was not found.
+        .ok_or_else(|| MolluskError::AccountIndexOverflow(key_map.len()))?;
```
buffalojoec: I don't think I understand this comment. First of all, I don't think `MolluskError<'static>` is the right lifetime. It should be `'a` and it should be shared with `&'a Instruction`. Then you could provide `instruction.program_id` to an error variant as a `&Pubkey`.

That being said, we should probably introduce a new error variant - say `ProgramIdNotMapped` - that looks just like one of these:
https://github.com/anza-xyz/mollusk/blob/a0d64e62197a7d2c6afc42d2b71e5471ad0fd997/error/src/error.rs#L21-L26

You could even make a case for `AccountMissing`.

buffalojoec: Also - what about calling `.map(|pos| u8::try_from(pos))` and unwrapping/panicking on that? This would catch `u8::MAX` violations earlier, similar to the location in the SDK linked in #179.

keys/src/accounts.rs
```diff
@@ -18,15 +18,39 @@ pub struct CompiledInstructionWithoutData {
...
+        .map(|account_meta| {
+            let index = key_map
+                .position(&account_meta.pubkey)
+                .ok_or_else(|| MolluskError::AccountIndexOverflow(key_map.len()))?;
+            if index > u8::MAX as usize {
+                Err(MolluskError::AccountIndexOverflow(index))
+            } else {
+                Ok(index as u8)
+            }
+        })
```
buffalojoec: `.map(..)` would probably simplify things here, too.

keys/src/accounts.rs
AvhiMaz: Thanks for the review, That comment was explaining a workaround that's no longer needed. The original code used a 'static lifetime which didn't allow borrowing references from the Instruction. So when a key wasn't found, we couldn't return the actual Pubkey in the error – we had to resort to returning AccountIndexOverflow(key_map.len()) as a magic value meaning "key not found".

I've fixed this by changing to use a borrowed lifetime 'a tied to the Instruction reference in the function signature. Now the return type uses MolluskError<'a> instead of MolluskError<'static>, which allows us to borrow pubkeys directly from the instruction.

keys/src/accounts.rs
```diff
@@ -15,18 +15,34 @@ pub struct CompiledInstructionWithoutData {
...
-            .collect(),
-    }
+    instruction: &'a Instruction,
+) -> Result<CompiledInstructionWithoutData, MolluskError<'a>> {
+    let program_id_index = key_map
+        .position(&instruction.program_id)
+        .ok_or(MolluskError::ProgramIdNotMapped(&instruction.program_id))
+        .and_then(|index| {
+            u8::try_from(index).map_err(|_| MolluskError::AccountIndexOverflow(index))
+        })?;
```
buffalojoec: Honestly, we were already using `.unwrap()` here, so I would rather just use `.or_panic_with` and not return the `Result`. What do you think?

https://github.com/anza-xyz/mollusk/blob/c77327d169dd7c02fa9662b5dfab1f7d730bb9ff/error/src/error.rs#L35-L37

You can see it used here:

https://github.com/anza-xyz/mollusk/blob/a0d64e62197a7d2c6afc42d2b71e5471ad0fd997/harness/src/file.rs#L64-L65

keys/src/accounts.rs
```diff
@@ -381,4 +397,91 @@ mod tests {
...
+
+        let instruction = Instruction::new_with_bytes(program_id, &[], vec![]);
+
+        match compile_instruction_without_data(&key_map, &instruction) {
+            Ok(_) => panic!("Expected overflow error, but got Ok"),
+            Err(MolluskError::AccountIndexOverflow(index)) => {
+                assert!(index > u8::MAX as usize);
+            }
+            Err(_) => panic!("Expected AccountIndexOverflow error"),
+        }
```
buffalojoec: On these tests you can just use `unwrap_err` and then either `assert_eq!` or `assert_matches!` to check the variant.

keys/src/accounts.rs
AvhiMaz: actually makes sense because the prev impl treated missing keys or index overflows as recoverable errors it returned a result with `Result<..., MolluskError>` suggesting that the caller might want to handle the error gracefully, using `or_panic_with` makes sense because these are validation errors not recoverable runtime conditions if an account is missing or an index overflows that indicates a test misconfiguration theres nothing meaningful to recover from so the test should panic immediately with a descriptive error, updated 👍

buffalojoec (approved):
Thanks!!

PR #182, opened 2025-11-29
harness: Add program with loader helper

febo:
### Problem

While there are a couple of `add_program*` helpers, there is none that adds a program under the `DEFAULT_LOADER_KEY`. This requires to always have to specify a loader when adding additional programs – see [here](https://github.com/febo/tide/blob/main/benchmark/benches/account.rs#L19-L37).

### Solution

Refactor the helpers to include 3 different variations:
* `add_program`
* `add_program_with_loader`
* `add_program_with_loader_and_elf`

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -655,23 +655,35 @@ impl Mollusk {
     /// - The current working directory
     pub fn new(program_id: &Pubkey, program_name: &str) -> Self {
         let mut mollusk = Self::default();
-        mollusk.add_program(program_id, program_name, &DEFAULT_LOADER_KEY);
+        mollusk.add_program(program_id, program_name);
         mollusk
     }
 
     /// Add a program to the test environment.
     ///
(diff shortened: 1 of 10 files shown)


harness/src/lib.rs
```diff
@@ -655,23 +655,35 @@ impl Mollusk {
...
-        self.add_program_with_elf_and_loader(program_id, &elf, loader_key);
+        self.add_program_with_loader_and_elf(program_id, &elf, loader_key);
     }
 
     /// Add a program to the test environment using a provided ELF under a
     /// specific loader.
     ///
     /// If you intend to CPI to a program, this is likely what you want to use.
-    pub fn add_program_with_elf_and_loader(
+    pub fn add_program_with_loader_and_elf(
```
sonicfromnewyoke, line 686: maybe it worth to change the order of arguments to the function also to not confuse naming and how to use?

febo: Good catch - updated.

buffalojoec (approved):

PR #184, opened 2025-12-05
feat: impl inner instructions tracking behind feature flag

### LINKED ISSUE #165: Inner instructions introspection
It would be great to inspect and assert what inner instructions were invoked. This is quite useful for [CPI events](https://www.anchor-lang.com/docs/features/events#emit_cpi) Anchor has.

Is this possible right now? If not, what changes would it require? Happy to contribute if you can provide some guidelines.

AvhiMaz:
Implement inner instructions (CPI) tracking behind feature flag.
Add `deconstruct_transaction` to extract CPIs from instruction trace.

Fixes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,10 +556,7 @@ This feature is particularly useful for:
 - Testing programs that make CPIs to verify correct invocation behavior
 - Debugging complex program interactions
 - Validating that the expected number of CPIs occur during execution
-
-Note: Inner instructions tracking adds a small overhead to program execution
-as it records each CPI. For performance-critical benchmarking where CPI
-tracking is not needed, you may want to disable this feature.
+- Inspecting the exact parameters passed to cross-program invocations
 
 ## Register tracing
 
--- a/harness/tests/bpf_program.rs
+++ b/harness/tests/bpf_program.rs
@@ -67,7 +67,6 @@ fn test_write_data() {
         &[(key, account.clone())],
         &[
             Check::success(),
-            #[cfg(not(feature = "inner-instructions"))]
             Check::compute_units(384),
             Check::account(&key)
                 .data(data)
@@ -153,7 +152,6 @@ fn test_transfer() {
         ],
         &[
             Check::success(),
-            #[cfg(not(feature = "inner-instructions"))]
             Check::compute_units(2480),
             Check::account(&payer)
                 .lamports(payer_lamports - transfer_amount)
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -94,6 +95,7 @@ solana-system-program = "3.1.0"
 solana-sysvar = "3.0"
 solana-sysvar-id = "3.0"
 solana-transaction-context = "3.1.0"
+solana-transaction-status-client-types = { version = "3.1.4", features = ["agave-unstable-api"] }
```
buffalojoec: nit: I try to specify only versions in the workspace manifest, and leave feature configurations up to the individual packages themselves.

harness/tests/bpf_program.rs
```diff
@@ -67,6 +67,7 @@ fn test_write_data() {
         &[(key, account.clone())],
         &[
             Check::success(),
+            #[cfg(not(feature = "inner-instructions"))]
```
buffalojoec: The inner instruction tracking has nothing to do with CUs. Are you using the Makefile to run the tests? It ensures you're using the same platform-tools.

harness/tests/bpf_program.rs
```diff
@@ -387,6 +391,109 @@ fn test_cpi() {
...
+        &[
+            (key, account.clone()),
+            (
+                cpi_target_program_id,
+                create_program_account_loader_v3(&cpi_target_program_id),
+            ),
+        ],
+        &[
+            Check::success(),
+            Check::inner_instruction_count(1),
```
buffalojoec, line 438: Cool - but maybe also write some validation of the inner instruction itself? You can pull it from the result.

README.md
```diff
@@ -499,6 +500,67 @@ for file in fs::read_dir(Path::new("fixtures-dir"))? {
...
+
+This feature is particularly useful for:
+- Testing programs that make CPIs to verify correct invocation behavior
+- Debugging complex program interactions
+- Validating that the expected number of CPIs occur during execution
+
+Note: Inner instructions tracking adds a small overhead to program execution
+as it records each CPI. For performance-critical benchmarking where CPI
+tracking is not needed, you may want to disable this feature.
+
```
buffalojoec: Hm, this really shouldn't be the case - otherwise validator replay wouldn't be deterministic. Where does this additional metering happen?

AvhiMaz: the issue i'm getting is because of mismatched versions of platform-tools, which created a huge confusion for me.

Issue #185, opened 2025-12-05
feature request: API to process multiple ixs with one `TransactionContext`


0xalpharush:
`process_instruction_chain` creates a new `TransactionContext` for each ix processed. This may make testing multi-ix behave differently than prod. in some cases e.g., calling `SyscallGetProcessedSiblingInstruction` syscall.

PR #186, opened 2025-12-06
feat: API to process multiple ixs with one TransactionContext

Duckaet:
it implemenst the issue #185


harness/src/compile_accounts.rs
```diff
@@ -63,3 +63,63 @@ where
...
+        /* program_id_index */ u16,
+        /* instruction_accounts */ Vec<InstructionAccount>,
+    )>,
+    pub transaction_accounts: Vec<(Pubkey, AccountSharedData)>,
+}
+
+pub fn compile_accounts_chain<'a, 'b>(
+    instructions: impl Iterator<Item = &'b Instruction>,
+    accounts: impl Iterator<Item = &'a (Pubkey, Account)>,
+    loader_key: Option<Pubkey>,
```
buffalojoec: We need to actually capture all owners of each program in the top-level instruction set, so we'll want to use something like a `Vec<(Pubkey, Pubkey)>` where we pass `(program_id, owner)`.

Then, we can stub only the top-level instructions' program accounts. What you have here won't work for a transaction context with two different program IDs invoked.

harness/tests/transaction_instructions.rs
buffalojoec, line 1: This test is a good start, but we'll want to really expand on the test coverage for the new API method. It would be good to run a few instructions against a few different programs in the same transaction context. We also should have at least one test that runs up against the transaction context limits, like number of accounts or CU budget.

harness/src/compile_accounts.rs
```diff
@@ -63,3 +63,63 @@ where
...
+    let transaction_accounts = compile_transaction_accounts(
+        &key_map,
+        &instruction_refs
+            .iter()
+            .map(|&ix| ix.clone())
+            .collect::<Vec<_>>(),
+        accounts,
+        Some(stub_out_program_account),
+        Some(fallback_to_instructions_sysvar),
+    );
```
buffalojoec: It feels like we have a lot of duplicated code here. I wonder if we can't just have a `compile_accounts` that does the compilation for an entire transaction context, and then call into that with just one instruction? Maybe even a slice with one element? That seems like the cleaner solution that writing all of this code over again for multiple elements.

harness/src/lib.rs
buffalojoec, line 1: I have a similar comment about duplicated code in this file. There's tons of duplication going on for the new API method you added. Let's see how we can share common code across all of the `process_` functions.

buffalojoec (commented):
Thanks for throwing this together! It's going to be a pretty hefty change - requiring lots of tests - so bear with me, but we can move toward a working solution.

Duckaet:
@buffalojoec , i tried to call compile_accounts but it broke some tests.. because they build different KeyMaps for single vs all instruction
```
KeyMap::compile(instruction_index, std::iter::once(instruction))
KeyMap::compile_for_chain(instructions.iter())
```

extracted out functions from process_transaction_instructions and validate

fixed the program owner bug, added more tests to verify the functionality

i have reduced duplicate code aswell with extract func

buffalojoec:
@Duckaet I just wrote a PR https://github.com/anza-xyz/mollusk/pull/187 that should make this change a lot easier. I refactored the way account fallbacks are handled during the account compilation step, which should enable you to share a `KeyMap` more easily.

harness/src/lib.rs
buffalojoec: I'm still seeing lots of code being duplicated, even after the rebase.

Duckaet: sorry for the delay,

as the process_ and validate were nearly identical with some checks, i implemented helpers for single,chain and multi fn

passed the core fn to inner aswell to reuse the code

is this valid? i couldnt think of a better fix to code duplication

harness/src/compile_accounts.rs
```diff
@@ -10,20 +10,23 @@ use {
     std::collections::{HashMap, HashSet},
 };
 
+/// Compile accounts for one or more instructions in a transaction context.
+///
+/// For single-instruction processing, pass a slice with one element:
+/// `&[instruction]`.
```
buffalojoec: nit: no need for this doc comment. The module is private.

harness/src/lib.rs
buffalojoec, line 1: Can't you just refactor `process_instruction_inner` to `process_transaction_instructions_inner`, take the same signature pattern you used in `compile_accounts`, and iterate over the instructions after `TransactionContext::new`?

harness/src/lib.rs
```diff
@@ -1219,6 +1381,176 @@ impl Mollusk {
...
+    /// JSON format.
+    ///
+    /// The `fuzz-fd` feature works the same way, but the variables require
+    /// the `_FD` suffix, in case both features are active together
+    /// (ie. `EJECT_FUZZ_FIXTURES_FD`). This will generate Firedancer fuzzing
+    /// fixtures, which are structured a bit differently than Mollusk's own
+    /// protobuf layouts.
+    pub fn process_and_validate_transaction_instructions(
+        &self,
+        instructions: &[(&Instruction, &[Check])],
```
buffalojoec: I know you used the API from `process_and_validate_instruction_chain` here, but since this is all supposed to be atomic, how about we just do one `checks` arg at the end?

harness/src/lib.rs
buffalojoec: This should get rid of most of the helpers you added.

PR #192, opened 2026-01-01
add file_name field to Mollusk struct

AkliluYirgalem:
groundwork for **multi-program benchmarking**: Added a _file_name_ field to **Mollusk** Struct. This allows the _bencher_ crate to track individual program filename, which i will use in the next PR to implement multi-program benchmarking and the filename will be used in reporting the Markdown file.

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -515,6 +515,7 @@ pub(crate) const DEFAULT_LOADER_KEY: Pubkey = solana_sdk_ids::bpf_loader_upgrade
 /// All fields can be manipulated through a handful of helper methods, but
 /// users can also directly access and modify them if they desire more control.
 pub struct Mollusk {
+    pub file_name: Option<String>,
     pub config: Config,
     pub compute_budget: ComputeBudget,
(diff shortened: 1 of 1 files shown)


buffalojoec (commented):
This relationship is a little misleading if we put the file name on the `Mollusk` struct. A `Mollusk` instance houses a program cache, which can hold many programs and be reused to invoke multiple programs at the top-level, especially over transaction contexts and instruction chains.

Is there a way we can refactor the bencher to support file name distinguishing instead?

buffalojoec:
> ```
> ProgramMatrixComputeUnitBencher::new()
>     .add_program(("program_v1", &mollust_v1))
>     .add_program(("program_v2", &mollust_v2))
>     .add_program(("program_v3", &mollust_v3))
>     .bench(("bench0", &instruction0, &accounts0))
>     .bench(("bench1", &instruction1, &accounts1))
>     .must_pass(true)
>     .out_dir("benches")
>     .execute();
> ```
>
> ...
> 
> what do you think?

OK sure, if you insist, let's give your way a try and see how it shakes out.

PR #193, opened 2026-01-08
Program cache to hold elf bytes

procdump:
**The problem**

`Mollusk` has the `register-tracing` feature. The way it reaches to the elf bytes is to walk the directories, search for the .so files and do the SHA-256 hashing onward. This can be skipped if we just keep the elf bytes in the program cache. Moreover this will come handy, I believe, when introducing the `sbpf-debugger` feature when we might need to map a program_id to elf bytes.

**What's changed**

We keep the elf_bytes per program_id in the program cache `Mollusk` maintains.
In regards to this enhancement and we can skip walking directories just to find the correct shared object, instead let's use the elf bytes from cache.

### DIFF
--- a/harness/src/file.rs
+++ b/harness/src/file.rs
@@ -27,7 +27,7 @@ use {
     },
 };
 
-pub(crate) fn default_shared_object_dirs() -> Vec<PathBuf> {
+fn default_shared_object_dirs() -> Vec<PathBuf> {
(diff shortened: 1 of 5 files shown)


buffalojoec (approved):
Yeah, it's about time we do this. I've already run into a few situations where this would have been useful.

Cleans up the register tracing nicely, too! Nice work!

procdump:
Great, thanks

PR #195, opened 2026-01-09
multiple program implementation bencher

AkliluYirgalem:
continuing the [discussion](https://github.com/anza-xyz/mollusk/pull/192) .

adds the MolluskComputeUnitMatrixBencher, allowing users to compare multiple Solana programs in a single execution.

Example Output:

  | Name     | CU (`program_v1`) | CU (`program_v2`) | CU (`program_v3`) |
  |----------|-------------------|-------------------|-------------------|
  | `bench0` | 1,400             | 1,390             | 1,385             |
 | `bench1` | 2,100             | 2,050             | 2,045             |


bencher/src/lib.rs
```diff
@@ -138,6 +141,96 @@ impl<'a> MolluskComputeUnitBencher<'a> {
     }
 }
 
+/// Mollusk's program matrix compute unit bencher.
+///
+/// Allows developers to bench test compute unit usage on multiple
+/// implementation of their programs.
```
buffalojoec:
```suggestion
/// implementations of their programs.
```

bencher/src/lib.rs
```diff
@@ -138,6 +141,96 @@ impl<'a> MolluskComputeUnitBencher<'a> {
...
+    }
+
+    /// Execute the benches.
+    pub fn execute(&mut self) {
+        let table_header = Utc::now().to_string();
+        let solana_version = get_solana_version();
+
+        let mut bench_results = Vec::new();
+
+        //Iterate through every instruction for every program
```
buffalojoec:
```suggestion
        // Iterate through every instruction for every program
```

bencher/src/result.rs
```diff
@@ -111,6 +128,76 @@ fn parse_last_md_table(content: &str) -> Vec<MolluskComputeUnitBenchResult<'_>>
...
+                mx_md_table.push_str("|\n");
+            }
+            // Start the new row with the instruction name
+            mx_md_table.push_str(&format!("| `{}` ", result.ix_name));
+            current_ix = result.ix_name;
+        }
+
+        // Add the CU value for this program in the current row
+        mx_md_table.push_str(&format!("| {} ", result.cus_consumed));
+    }
```
buffalojoec: Instead of all this guesswork, why don't you just have the function accept a matrix?
You can probably use the existing result type as well.

```rust
Vec<Vec<(String, MolluskComputeUnitBenchResult)>>
```

bencher/src/result.rs
```diff
@@ -111,6 +128,76 @@ fn parse_last_md_table(content: &str) -> Vec<MolluskComputeUnitBenchResult<'_>>
...
+    let path = out_dir.join("mx_compute_units.md");
+    prepend_to_md_file(&path, &mx_md_table);
+}
+
+fn mx_md_header(
+    table_header: &str,
+    solana_version: &str,
+    results: &[MolluskComputeUnitMatrixBenchResult],
+) -> String {
+    //Get ONLY the unique program names.
```
buffalojoec: I'm surprised fmt doesnt catch these comment issues.

bencher/tests/markdown.rs
```diff
@@ -27,3 +29,34 @@ fn test_markdown() {
...
+
+    //this stinks, but should be enough for demonstration.
+    let mollusk1 = Mollusk::new(&program_id, "test_program_primary");
+    let mollusk2 = Mollusk::new(&program_id, "test_program_primary");
+    let mollusk3 = Mollusk::new(&program_id, "test_program_primary");
+
+    MolluskComputeUnitMatrixBencher::new()
+        .add_program(("program_v1", &mollusk1))
+        .add_program(("program_v2", &mollusk2))
+        .add_program(("program_v3", &mollusk3))
```
buffalojoec: This is where I think the API looks a bit funny. You have to provide the program name twice. That was why I was suggesting tucking the `Mollusk` instance under the hood and just hydrating it from the bencher's API.

bencher/tests/markdown.rs
buffalojoec, line 1: Can we get an example output in the README and the PR description?

bencher/src/result.rs
AkliluYirgalem: > Instead of all this guesswork, why don't you just have the function accept a matrix? You can probably use the existing result type as well.
> 
> ```rust
> Vec<Vec<(String, MolluskComputeUnitBenchResult)>>
> ```

this is actually pretty nice, i felt itchy writing that function, but, wouldn't the `String` be redundant? maybe something like `Vec<Vec<MolluskComputeUnitBenchResult>>`?

buffalojoec: You're right, the string is too much. It's better to do an ID for the column key, but if you can get away without it, sure.

bencher/tests/markdown.rs
AkliluYirgalem: sure, then i should use the internal `add_program` of `Mollusk`, and just load the program inside the bencher from a given list of program names.

test-programs/noop-log/src/lib.rs
buffalojoec, line 1: Is this program's implementation any different from the v1? If not, how about just adding a line in the Makefile to copy the `primary` binary instead of duplicating the source?

Cargo.toml
```diff
@@ -72,6 +72,7 @@ solana-loader-v4-interface = "3.1.0"
 solana-loader-v4-program = "3.1.0"
 solana-logger = "3.0"
 solana-message = "3.0.1"
+solana-msg = "3.0.0"
```
buffalojoec, line 75: Is this needed?

AkliluYirgalem: Yes, its used to add compute usage for the `primary_v2` program, which just logs "Instruction: 0".

test-programs/noop-log/src/lib.rs
AkliluYirgalem: its different, it intentionally consumes more compute than the v1 by logging a string.

AkliluYirgalem:
> This is looking great! We just need to figure out a better way to test two test programs, then I think we're good to go!

i have already added the `primary_v2`, what do you suggest?

test-programs/noop-log/src/lib.rs
buffalojoec: Does it need to have all five of the same instruction processors? Can we simplify it?

AkliluYirgalem: nope, removed it.

README.md
buffalojoec, line 1: Can you wrap your markdown changes to line width 80 (besides code snippets)?

test-programs/noop-log/Cargo.toml
buffalojoec, line 1: Can you rename this program from `primary_v2` to `noop_log`?

PR #196, opened 2026-02-01
Fix/sysvars in fixture

0xalpharush:
Fixtures exported from mollusk were triggering an [unwrap](https://github.com/firedancer-io/solfuzz-agave/blob/3e12a4b269083d16b918bc8885c89fa6f6b556d7/src/lib.rs#L793) in solfuzz-agave because sysvars not referenced in the transaction accounts were not included and the `RecentBlockhashes` sysvar was unimplemented. I've fixed these two issues.

### DIFF
--- a/fuzz/fixture/Cargo.toml
+++ b/fuzz/fixture/Cargo.toml
@@ -18,7 +18,7 @@ solana-clock = { workspace = true }
 solana-compute-budget = { workspace = true, features = ["agave-unstable-api"] }
 solana-epoch-rewards = { workspace = true }
 solana-epoch-schedule = { workspace = true }
-solana-fee-calculator = "3.0.0"
+solana-fee-calculator = { workspace = true }
 solana-hash = { workspace = true }
 solana-instruction = { workspace = true }
 solana-keccak-hasher = { workspace = true }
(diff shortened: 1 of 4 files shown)


fuzz/fixture/Cargo.toml
```diff
@@ -18,13 +18,15 @@ solana-clock = { workspace = true }
 solana-compute-budget = { workspace = true, features = ["agave-unstable-api"] }
 solana-epoch-rewards = { workspace = true }
 solana-epoch-schedule = { workspace = true }
+solana-fee-calculator = "3.0.0"
```
buffalojoec: Can you declare this at the workspace level like the rest of the dependencies?

harness/src/sysvar.rs
buffalojoec, line 1: We should probably also add to `setup_sysvar_cache`, right?

buffalojoec: Same with `From<&Sysvars> for SysvarCache`.

fuzz/fixture/src/sysvars.rs
buffalojoec, line 1: Can you add to `hash_proto_sysvars`?

buffalojoec (approved):

PR #199, opened 2026-02-16
feat: Dump a register trace disassembly if SBF_TRACE_DISASSEMBLE is set

procdump:
### What's changed

Dump a register trace disassembly if requested by user taking advantage of the API in `solana-sbpf`.

### DIFF
--- a/harness/src/register_tracing.rs
+++ b/harness/src/register_tracing.rs
@@ -37,7 +37,7 @@ impl DefaultRegisterTracingCallback {
         ) {
             Ok(analysis) => {
                 if let Err(e) = analysis.disassemble_register_trace(writer, register_trace) {
-                    eprintln!("Can't dissasemble register trace for {program_id}: {e:#?}");
+                    eprintln!("Can't disassemble register trace for {program_id}: {e:#?}");
                 }
             }
             Err(e) => {
@@ -73,7 +73,7 @@ impl DefaultRegisterTracingCallback {
 
         // Persist a full trace disassembly if requested.
         if self.sbf_trace_disassemble {
-            let mut trace_disassemble_file = File::create(base_fname.with_extension("disassem"))?;
+            let mut trace_disassemble_file = File::create(base_fname.with_extension("trace"))?;
             self.disassemble_register_trace(
                 &mut trace_disassemble_file,
                 program_id,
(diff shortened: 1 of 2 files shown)


harness/src/register_tracing.rs
```diff
@@ -11,18 +11,41 @@ const DEFAULT_PATH: &str = "target/sbf/trace";
...
+        program_id: &Pubkey,
+        executable: &Executable,
+        register_trace: RegisterTrace,
+    ) {
+        match solana_program_runtime::solana_sbpf::static_analysis::Analysis::from_executable(
+            executable,
+        ) {
+            Ok(analysis) => {
+                if let Err(e) = analysis.disassemble_register_trace(writer, register_trace) {
+                    eprintln!("Can't dissasemble register trace for {program_id}: {e:#?}");
```
buffalojoec:
```suggestion
                    eprintln!("Can't disassemble register trace for {program_id}: {e:#?}");
```

harness/src/register_tracing.rs
```diff
@@ -47,6 +70,18 @@ impl DefaultRegisterTracingCallback {
 
         // Get program_id.
         let program_id = instruction_context.get_program_key()?;
+
+        // Persist a full trace disassembly if requested.
+        if self.sbf_trace_disassemble {
+            let mut trace_disassemble_file = File::create(base_fname.with_extension("disassem"))?;
```
buffalojoec: I'm not sure the best file extension here, but what about `.trace`?

procdump:
> @procdump this is a breaking change, so do you want me to wait to publish until you've landed any more register tracing changes?

I'm currently preparing the `sbpf-debugger` feature contribution. It also has some breaking changes to the API, yet I can't fully submit it because mollusk is still tied to 3.1.x and the necessary debugger PRs actually landed in agave master 4.x. However, I can split the debugger feature into smaller ones moving out code that also pertain to the register tracing changes. Does this sound reasonable to you? With this approach we can reduce the breaking versions I believe.

buffalojoec:
> I'm currently preparing the `sbpf-debugger` feature contribution. It also has some breaking changes to the API, yet I can't fully submit it because mollusk is still tied to 3.1.x and the necessary debugger PRs actually landed in agave master 4.x. However, I can split the debugger feature into smaller ones moving out code that also pertain to the register tracing changes. Does this sound reasonable to you? With this approach we can reduce the breaking versions I believe.

Yeah, I don't mind many subsequent changes, but I'd prefer to release once we have all of them in. I can release earlier if you need to use this change, though.

PR #200, opened 2026-02-24
SBPF debugger integration and expression-based trace filtering

procdump:
## The problem

Mollusk's register tracing had no way to selectively trace or debug specific programs — it was all-or-nothing. There was also no integration with the SBPF debugger, meaning users couldn't interactively debug program execution from within Mollusk tests.

## What's changed

- **Expression-based trace filtering** (`SBF_TRACE_FILTER`) — filter which programs produce traces using boolean expressions (e.g. `program_id == <pubkey>`), built with `nom`. Currently only `program_id` is exposed as a filterable field, but the parser and evaluator are generic — adding more fields is straightforward as long as it's aligned with the behavior of the runtime.
- **SBPF debugger integration** (`SBF_DEBUG_PORT`, `sbpf-debugger` feature) — starts a GDB remote stub when the filter matches, allowing interactive debugging over TCP. Reuses the register tracing callback to avoid duplicating filtering logic. Separate feature because it forces interpreter mode.
- **ELF SHA-256 mapping file** — persists `program_id → ELF hash` for deterministic `.so` resolution by debugger clients.
- **Runtime callback downcasting** (`as_any`/`as_any_mut`) — allows modifying callback fields at runtime without rebuilding Mollusk.
- **Tests** — unit tests for the filter expression parser/evaluator, and an integration test for the debugger that verifies filter-based activation (both match and no-match), metadata exchange over the GDB protocol, and the ELF SHA-256 mapping table.
- Vendored `solana-program-runtime`, `solana-bpf-loader-program`, and `solana-sbpf`, etc.. to backport required upstream changes (temporary, will be removed when mollusk aligns with agave 4.x). Please, ignore the `vendor` directory.

See [MOTIVATION.md](https://github.com/procdump/mollusk/blob/sbpf-debugger/MOTIVATION.md) for detailed rationale behind each change.

Note: CI fails due to the vendoring of sources which is temporary.

### DIFF
--- a/harness/src/register_tracing_filter.rs
+++ b/harness/src/register_tracing_filter.rs
@@ -35,6 +35,7 @@ pub enum Expr<'a> {
 
 // --- Parser ---
 
+/// Wraps a parser to strip leading and trailing whitespace.
 fn ws<'a, F, O>(inner: F) -> impl Parser<&'a str, Output = O, Error = nom::error::Error<&'a str>>
 where
     F: Parser<&'a str, Output = O, Error = nom::error::Error<&'a str>>,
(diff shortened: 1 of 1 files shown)


procdump:
As https://github.com/anza-xyz/mollusk/pull/229 got merged, I'm closing PR. The filtering part can be extracted in a follow up later.

PR #201, opened 2026-03-02
fix: prevent duplicate sysvar accounts and ensure sysvars in fixtures

0xalpharush:
Two bugs in `load_accounts_for_instructions`:

1. `seen` was initialized empty after sysvars were already pushed into `accounts`, so instruction-referenced sysvars were added a second time, triggering a "duplicate program account" invariant violation. Fix: seed `seen` from the already-collected accounts.

2. Sysvars were only added when `hydrate_store` was enabled. Mollusk execution worked (sysvar cache is populated separately), but serialized fixtures couldn't find sysvar accounts. Fix: add sysvars unconditionally, guarded only by whether the store already provides an override.

These bugs were causing exported instr fixtures to fail in solfuzz-agave .

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -1792,35 +1792,37 @@ impl<AS: AccountStore> MolluskContext<AS> {
                 .program_cache
                 .get_all_keyed_program_accounts()
                 .into_iter()
-                .chain(self.mollusk.sysvars.get_all_keyed_sysvar_accounts())
                 .for_each(|(pubkey, account)| {
                     accounts.push((pubkey, account));
                 });
         }
+        let store = self.account_store.borrow();
+        self.mollusk
+            .sysvars
+            .get_all_keyed_sysvar_accounts()
+            .into_iter()
+            .for_each(|(pubkey, account)| {
+                if store.get_account(&pubkey).is_none() {
+                    accounts.push((pubkey, account));
+                }
+            });
 
         // Regardless of hydration, only add an account if the caller hasn't
         // already loaded it into the store.
-        let mut seen = HashSet::new();
-        let store = self.account_store.borrow();
+        let mut seen: HashSet<Pubkey> = accounts.iter().map(|(k, _)| *k).collect();
         instructions.for_each(|instruction| {
             instruction
                 .accounts
                 .iter()
                 .for_each(|AccountMeta { pubkey, .. }| {
                     if seen.insert(*pubkey) && pubkey != &solana_instructions_sysvar::id() {
-                        // First try to load theirs, then see if it's a sysvar,
-                        // then see if it's a cached program, then apply the
-                        // default.
+                        // First try to load theirs, then see if it's a cached program, then apply
+                        // the default.
                         let account = store.get_account(pubkey).unwrap_or_else(|| {
                             self.mollusk
-                                .sysvars
-                                .maybe_create_sysvar_account(pubkey)
-                                .unwrap_or_else(|| {
-                                    self.mollusk
-                                        .program_cache
-                                        .maybe_create_program_account(pubkey)
-                                        .unwrap_or_else(|| store.default_account(pubkey))
-                                })
+                                .program_cache
+                                .maybe_create_program_account(pubkey)
+                                .unwrap_or_else(|| store.default_account(pubkey))
                         });
                         accounts.push((*pubkey, account));
                     }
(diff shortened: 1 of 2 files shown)


harness/src/lib.rs
```diff
@@ -1792,35 +1792,37 @@ impl<AS: AccountStore> MolluskContext<AS> {
...
         }
+        let store = self.account_store.borrow();
+        self.mollusk
+            .sysvars
+            .get_all_keyed_sysvar_accounts()
+            .into_iter()
+            .for_each(|(pubkey, account)| {
+                if store.get_account(&pubkey).is_none() {
+                    accounts.push((pubkey, account));
+                }
```
buffalojoec: I think it's probably better to just do this on line 1796 (which means it would apply to programs and sysvars) and then update the comment on line 1788-1789.

harness/src/lib.rs
buffalojoec, line 1790: By design this flag is supposed to force hydration of the store with everything that's current on the `Mollusk` instance. We shouldn't cut sysvars out of that rule.

buffalojoec, line 1753: When writing unit tests, many devs may use `MolluskContext` mutably and directly manipulate the program cache and sysvars. The idea behind the `hydrate_store` flag is to essentially sync the execution with those changes.

That being said, we probably don't want to make this the default without a way to opt-out, especially in the `fuzz`. What about adding a config parameter to `with_context`?

```rust
pub fn with_context<AS: AccountStore>(
  self,
  mut account_store: AS,
  config: ContextConfig,
) -> MolluskContext<AS> {
  ...
  MolluskContext {
    ...
    hydrate_store: config.hydrate_store,
  }
}

pub struct ContextConfig {
  hydrate_store: bool,
}
```

harness/src/lib.rs
```diff
@@ -1792,35 +1792,37 @@ impl<AS: AccountStore> MolluskContext<AS> {
...
                         let account = store.get_account(pubkey).unwrap_or_else(|| {
                             self.mollusk
-                                .sysvars
-                                .maybe_create_sysvar_account(pubkey)
-                                .unwrap_or_else(|| {
-                                    self.mollusk
-                                        .program_cache
-                                        .maybe_create_program_account(pubkey)
-                                        .unwrap_or_else(|| store.default_account(pubkey))
-                                })
```
buffalojoec, line 1823: If hydration wasn't forced, do we still need these changes? Seems like we could just disable force-hydration in the fuzz API and this whole problem goes away?

harness/src/lib.rs
0xalpharush: The code was moved out of the branch so it now always adds the sysvar. It prioritizes the `Mollusk` instance version if it exists by using `get_account`

buffalojoec: The code that you wrote for sysvars only adds one if the account doesn't exist already. The feature was designed for force-adding (overwriting).

harness/src/lib.rs
0xalpharush: I am using the following with `EJECT_*` env
```
    let mollusk_context = MolluskContext {
        mollusk,
        account_store: Rc::new(RefCell::new(accounts)),
        hydrate_store: false,
    };
    let result = mollusk_context.process_instruction(&instruction);
```
This fallback is redundant to have as they are already added whether the instruction references it https://github.com/anza-xyz/mollusk/pull/201#discussion_r2887300792

harness/src/lib.rs
```diff
@@ -1792,35 +1792,37 @@ impl<AS: AccountStore> MolluskContext<AS> {
...
+        let store = self.account_store.borrow();
+        self.mollusk
+            .sysvars
+            .get_all_keyed_sysvar_accounts()
+            .into_iter()
+            .for_each(|(pubkey, account)| {
+                if store.get_account(&pubkey).is_none() {
+                    accounts.push((pubkey, account));
+                }
+            });
```
buffalojoec, line 1808: Ah okay, I think I see the problem (re: [this comment](https://github.com/anza-xyz/mollusk/pull/201/changes#r2887311048)). What if instead we just did this on line 1796?

```rust
.for_each|(pubkey, account)| {
  if hydrate_store {
    // Hydrate: force-update the store.
    accounts.push((pubkey, account));
  } else {
    // Only add those that don't exist.
    if store.get_account(&pubkey).is_none() {
      accounts.push((pubkey, account));
    }
  }
})
```

and then just ran this loop always? Then I think below this, the code can go back to the way it was, right?

buffalojoec: I'm basically trying to give program accounts the same treatment you're giving sysvars in this change.

PR #202, opened 2026-03-05
cli: (fix) ignore compute units flag check override

puhtaytow:
### Issue

As per flag description `ignore_compute_units` when set, should remove `ComputeUnits` from checks list, which it didn't in case of provided with config.

### Summary of changes
- Small refactor around checks loading and fixed the issue
- Added test

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -145,38 +145,6 @@ fn add_elf_to_mollusk(mollusk: &mut Mollusk, elf_path: &str, program_id: &Pubkey
     );
 }
 
-fn resolve_checks(
-    checks_from_config: Option<Vec<Compare>>,
-    ignore_compute_units: bool,
-) -> Vec<Compare> {
-    let mut checks = if let Some(config_checks) = checks_from_config {
-        config_checks
-    } else if ignore_compute_units {
-        // Defaults to all checks except compute units when requested.
-        Compare::everything_but_cus()
-    } else {
-        // Defaults to all checks.
-        Compare::everything()
-    };
-
-    // Ditch ComputeUnits checks to reflect CU flag override.
-    if ignore_compute_units {
-        checks.retain(|check| !matches!(check, Compare::ComputeUnits));
-    }
-
-    checks
-}
-
-fn set_checks(
-    config: Option<String>,
-    ignore_compute_units: bool,
-) -> Result<Vec<Compare>, Box<dyn std::error::Error>> {
-    let checks_from_config = config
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -164,15 +196,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
             add_elf_to_mollusk(&mut mollusk, &elf_path, &program_id);
 
-            let checks = if let Some(config_path) = config {
-                ConfigFile::try_load(&config_path)?.checks
-            } else if ignore_compute_units {
-                Compare::everything_but_cus()
-            } else {
-                // Defaults to all checks.
-                Compare::everything()
-            };
```
buffalojoec, line 174: Weird, it certainly seems like it should be working, and we've used the CLI with CUs ignored before.
https://github.com/anza-xyz/mollusk/blob/16e95bc3c83956f923ea044e472e5d8a0ba88664/result/src/compare.rs#L130

buffalojoec (commented):
Can you maybe describe the issue more? I know we've used this feature without issue before.

cli/src/main.rs
puhtaytow: > Can you maybe describe the issue more? 

To my understanding,  the `--ignore-compute-units` should reflect in guaranteed skip of the CU checks even if these are present in config, as reference this [comment](https://github.com/anza-xyz/mollusk/blob/main/cli/src/main.rs#L95C1-L98C36)

The current implementation picks  the config set (if present) and skips the override flag, which could appear to work if the config did not include CU checks, i think it could also remain undetected in case of having zero change to the compared CU values.

buffalojoec: Ohh I see, so the main issue is that it doesn't override the config, but you want to be able to still use all other values provided by the config?

puhtaytow: > Ohh I see, so the main issue is that it doesn't override the config, but you want to be able to still use all other values provided by the config?

Exactly :)

buffalojoec: Can't you just do this in-line?

```rust
let checks = if let Some(config_path) = config {
  let mut checks = ConfigFile::try_load(&config_path)?.checks;
  if ignore_compute_units {
    checks.retain(|check| !matches!(check, Compare::ComputeUnits));
  }
  checks
} else if ignore_compute_units {
  Compare::everything_but_cus()
} else {
  // Defaults to all checks.
  Compare::everything()
};
```

puhtaytow: Surely, if you prefer that way. I had it extracted to separate function as the original block was duplicated code.

PR #203, opened 2026-03-10
Update to 4.0 agave deps

grod220:
Updating to use new agave 4.0 deps

### DIFF
--- a/harness/tests/bpf_program.rs
+++ b/harness/tests/bpf_program.rs
@@ -152,7 +152,7 @@ fn test_transfer() {
         ],
         &[
             Check::success(),
-            Check::compute_units(2484),
+            Check::compute_units(2498),
             Check::account(&payer)
                 .lamports(payer_lamports - transfer_amount)
                 .build(),
@@ -256,7 +256,7 @@ fn test_close_account() {
         ],
         &[
             Check::success(),
-            Check::compute_units(2560),
+            Check::compute_units(2574),
             Check::account(&key)
                 .closed() // The rest is unnecessary, just testing.
                 .data(&[])
@@ -376,7 +376,7 @@ fn test_cpi() {
         ],
         &[
             Check::success(),
-            Check::compute_units(2326),
+            Check::compute_units(2341),
             Check::account(&key)
                 .data(data)
                 .lamports(lamports)


grod220:
@buffalojoec any thoughts about creating a 4.0 deps branch in this repo and shipping "0.13.0-beta.1" to crates.io so consumers like solana-stake-interface/program can depend on that versus a git dep?

buffalojoec:
> @buffalojoec any thoughts about creating a 4.0 deps branch in this repo and shipping "0.13.0-beta.1" to crates.io so consumers like solana-stake-interface/program can depend on that versus a git dep?

Sure! Done: https://github.com/anza-xyz/mollusk/tree/agave-4.0

.github/workflows/main.yml
buffalojoec, line 1: Nice, thanks for grabbing this!

harness/tests/bpf_program.rs
```diff
@@ -152,7 +152,7 @@ fn test_transfer() {
         ],
         &[
             Check::success(),
-            Check::compute_units(2482),
+            Check::compute_units(2484),
```
buffalojoec: Which version of the CLI did you use for this branch? Can we update the Makefile's `SOLANA_VERSION`?

Cargo.toml
```diff
@@ -80,27 +80,27 @@ solana-precompile-error = "3.0"
...
 solana-slot-hashes = "3.0"
-solana-stake-interface = "2.0.1"
-solana-svm-callback = "3.1.0"
-solana-svm-log-collector = "3.1.0"
-solana-svm-timings = "3.1.0"
-solana-svm-transaction = "3.1"
-solana-system-interface = "2.0"
-solana-system-program = "3.1.0"
-solana-sysvar = "3.1"
+solana-stake-interface = { version = "3.1.0", features = ["bincode", "sysvar"] }
```
buffalojoec: I'd rather define which features to enable in the crates themselves who use them, rather than the workspace manifest. Can you move this and the ones on `solana-sysvar` into whichever manifest needs them?

harness/src/lib.rs
```diff
@@ -896,36 +897,95 @@ impl Mollusk {
...
+                .configure_top_level_instruction_for_tests(program_index, Vec::new(), Vec::new())
+                .expect("failed to seed top-level instruction index");
+            transaction_context
+                .push()
+                .expect("failed to advance top-level instruction index");
+            transaction_context
+                .pop()
+                .expect("failed to reset seeded instruction frame");
+        }
+    }
```
buffalojoec: We shouldn't need this. Mollusk already uses a `SantizedMessage` under the hood for all of its instruction processor methods, so we should be able to just call into [`TransactionContext::prepare_top_level_instructions`](https://github.com/anza-xyz/agave/blob/bf440752b4b2dd468af19b0acb14aa4c21b99590/program-runtime/src/invoke_context.rs#L504).

You can see how Agave does it in SVM's [message processor](https://github.com/anza-xyz/agave/blob/bf440752b4b2dd468af19b0acb14aa4c21b99590/svm/src/message_processor.rs#L20-L26).

harness/src/lib.rs
```diff
@@ -896,36 +897,95 @@ impl Mollusk {
...
+
+    // TODO: Agave 4's transaction context still consumes the older account/rent
```
buffalojoec, line 932: This is because Mollusk is using newer `solana-rent` than Agave is? Can we tweak the deps on this branch to align with Agave? Then we don't need any of the versioned dependencies like `solana-account-legacy` and `solana-rent-legacy`.

harness/tests/bpf_program.rs
grod220: Updated to the latest 👍

Cargo.toml
grod220: Makes sense. Fixed ✅

harness/src/lib.rs
grod220: Oh! That's indeed helpful, but it's not available in `v4.0.0-beta.6`. I added a comment to replace this entirely on the next version bump.

harness/src/lib.rs
grod220: It would make it easier for mollusk, but would push this incompatibility issue downstream. Consumers like the stake program are wanting to use the new v4 types, but the mollusk API would be behind.

PR #204, opened 2026-03-12
perf: add cached program runtime env

sonicfromnewyoke:
### Problem

 `create_program_runtime_environment_v1`/`v2` called on every `process_transaction_message` invocation, rebuilding the entire syscall registry each time

### Proposed Solution

add cache program runtime environments with 2 function: for build and rebuild

#### Benches

| Test.                           | Before (us) | After (us) | Improvement     |
| -------------------- | ------------ | ---------- | --------------- |
| transfer_checked     | 38.105         | 33.431       |  ~12.3%.          |
| transfer_unchecked | 38.007         | 34.554       | ~ 9.1%             |

before:
```
transfers/transfer_checked
                        time:   [38.053 µs 38.105 µs 38.159 µs]
                        thrpt:  [26.206 Kelem/s 26.243 Kelem/s 26.279 Kelem/s]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) low mild

transfers/transfer_unchecked
                        time:   [37.927 µs 38.007 µs 38.093 µs]
                        thrpt:  [26.251 Kelem/s 26.311 Kelem/s 26.367 Kelem/s]
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
 ```
 
 after:
```
transfers/transfer_checked
                        time:   [33.361 µs 33.431 µs 33.503 µs]
                        thrpt:  [29.848 Kelem/s 29.913 Kelem/s 29.975 Kelem/s]
                 change:
                        time:   [−4.2487% −3.9360% −3.6447%] (p = 0.00 < 0.05)
                        thrpt:  [+3.7826% +4.0973% +4.4373%]
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild

transfers/transfer_unchecked
                        time:   [34.288 µs 34.554 µs 34.843 µs]
                        thrpt:  [28.700 Kelem/s 28.940 Kelem/s 29.164 Kelem/s]
                 change:
                        time:   [−1.7958% −1.0694% −0.0471%] (p = 0.01 < 0.05)
                        thrpt:  [+0.0471% +1.0809% +1.8287%]
                        Change within noise threshold.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2067,9 +2067,9 @@ dependencies = [
 
 [[package]]
 name = "keccak"
-version = "0.1.5"
+version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ecc2af9a1119c51f12a14607e783cb977bde58bc069ff0c3da1095e635d70654"
+checksum = "cb26cec98cce3a3d96cbb7bced3c4b16e3d13f27ec56dbd62cbc8f39cfb9d653"
 dependencies = [
  "cpufeatures",
 ]
@@ -2913,9 +2913,9 @@ dependencies = [
 
 [[package]]
 name = "quinn-proto"
-version = "0.11.13"
+version = "0.11.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1906b49b0c3bc04b5fe5d86a77925ae6524a19b816ae38ce1e426255f1d8a31"
+checksum = "434b42fec591c96ef50e21e886936e66d3cc3f737104fdb9b737c40ffb94c098"
 dependencies = [
  "bytes",
  "getrandom 0.3.4",


buffalojoec (commented):
Very nice, and thanks for the stats!

This is actually a very tricky thing to get right, since the program runtime environment (PRE) is schlepped around the validator. Essentially, there are two main areas we need to consider.

When a program is deployed and its ELF is loaded, the PRE serves as the loader which feeds ELF parsing and relocations. I think maybe here we can just always use `ComputeBudget::default` and `FeatureSet::all_enabled`. Programs are loaded into the test environment's cache manually, so there's no need to try to toggle ELF loading errors.
https://github.com/anza-xyz/agave/blob/3d720d2ca01783fc71dcbb40a7192607e3bc1cf2/program-runtime/src/deploy.rs#L148
https://github.com/anza-xyz/agave/blob/3d720d2ca01783fc71dcbb40a7192607e3bc1cf2/program-runtime/src/deploy.rs#L77-L80
https://github.com/anza-xyz/sbpf/blob/31a4b36f7eee5385e872592eb28e6bceb5ce7cfd/src/elf.rs#L692-L697

The PRE we really care about is the *execution* environment. This is the one that will throw a transaction error if a program tries to use a syscall whose feature is inactive, exceed the max CPI depth, etc. This one we definitely want people to be able to play with.
https://github.com/anza-xyz/agave/blob/3d720d2ca01783fc71dcbb40a7192607e3bc1cf2/program-runtime/src/invoke_context.rs#L588-L598

The way I see it, we can do one of two things:
* Stop allowing `ComputeBudget` and `FeatureSet` to be directly manipulated and keep syncing the PRE under the hood as you are here
* Roll something custom to cache as much as possible during `Mollusk` instantiation

The something custom I have in mind is to basically cache all of the `Config` struct except for the two fields that come from `ComputeBudget` as well as all of the syscalls that are not feature-gated.
https://github.com/anza-xyz/agave/blob/3d720d2ca01783fc71dcbb40a7192607e3bc1cf2/syscalls/src/lib.rs#L329-L346

Then, during execution, we check only the feature-gated syscalls and optionally register them before invoking

What do you think? I'm actually happy to upstream this to Agave if it's what we decide to do.

buffalojoec:
@sonicfromnewyoke I'm actually going to cook on something on the Agave side for a bit and see if I can come up with a good way to do this, and then we can probably take inspiration from there.

Thanks for raising this issue. I'll loop back here ASAP when I'm done!

PR #206, opened 2026-03-17
cli: preloading extra programs for fixtures

grod220:
Adds three new repeatable CLI flags:
- --add-program
- --add-program-with-loader
- --add-program-with-loader-and-elf

This allows users to preload additional programs (e.g. SPL Token) into the Mollusk SVM runtime without modifying test fixtures.

Needed in p-ATA fixture work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ target/
 .DS_Store
 
 harness/tests/test-vectors/
+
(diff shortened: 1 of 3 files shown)


buffalojoec (approved):
Awesome work! Clean implementation and useful feature.

PR #209, opened 2026-03-23
adding trusted publishing + SLSA provenance

nbelenkov:
Replacing the `CRATES_IO_TOKEN` secret with OIDC-based trusted publishing via `rust-lang/crates-io-auth-action`. This eliminates the need for a long-lived API token in repo secrets.

**Changes:**

- Add `rust-lang/crates-io-auth-action@v1` to obtain a short-lived OIDC token at publish time
- Gate the publish job on a prod environment for an additional approval layer
- Add SLSA build provenance attestation for all published .crate files via `actions/attest-build-provenance@v4`
- Add `permissions: {}` at workflow level and explicit per-job permission blocks
- Add `permissions: contents: read` to pre_publish (required after tightening workflow-level defaults) 

**Required setup before merging:**

- Create a prod environment in GitHub repo settings
- Add a trusted publisher on crates.io for each of the 10 crates (owner: anza-xyz, repo: mollusk, workflow: publish.yml, environment: prod)
- Delete the `CRATES_IO_TOKEN` secret once the new workflow is confirmed working

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -73,7 +73,7 @@ jobs:
         with:
           cache: true
           version: ${{ steps.solana.outputs.version }}
-      - uses: rust-lang/crates-io-auth-action@v1
+      - uses: rust-lang/crates-io-auth-action@b7e9a28eded4986ec6b1fa40eeee8f8f165559ec  # v1.0.3
         id: auth
       - name: Publish
         shell: bash
@@ -82,6 +82,6 @@ jobs:
           ARGS: "--locked"
         run: make publish
       - name: Attest build provenance
-        uses: actions/attest-build-provenance@v4
+        uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32  # v4.1.0
         with:
           subject-path: target/package/*.crate


.github/workflows/publish.yml
```diff
@@ -43,8 +47,14 @@ jobs:
   publish:
     name: Publish
     runs-on: ubuntu-latest
+    environment: prod
     needs: pre_publish
     if: github.ref == 'refs/heads/main'
+    permissions:
+      id-token: write
+      attestations: write
+      artifact-metadata: write
```
buffalojoec, line 56: What's this one for?

.github/workflows/publish.yml
```diff
@@ -63,9 +73,15 @@ jobs:
         with:
           cache: true
           version: ${{ steps.solana.outputs.version }}
+      - uses: rust-lang/crates-io-auth-action@v1
+        id: auth
```
buffalojoec: How long is this token good for? The publishing run takes a bit, since it does all the libraries in serial.

.github/workflows/publish.yml
```diff
@@ -63,9 +73,15 @@ jobs:
...
       - name: Publish
         shell: bash
         env:
-          TOKEN: ${{ secrets.CRATES_IO_TOKEN }}
+          TOKEN: ${{ steps.auth.outputs.token }}
           ARGS: "--locked"
         run: make publish
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@v4
```
buffalojoec: Some extra security for this and `rust-lang/crates-io-auth-action` would be to use an actual git SHA instead of a tag, so we can treat the digest as immutable. What do you think?

.github/workflows/publish.yml
nbelenkov: 30 mins, I saw in action last one took about 6 mins to publish so we should be good

.github/workflows/publish.yml
nbelenkov: Yeah, that is the correct way to do this. But we don't have Dependabot in the repo atm. So my plan was to add Dependabot in a new PR for actions and then change to hashes that can be bumped.

.github/workflows/publish.yml
nbelenkov: This one is needed to upload the attestation to the github attestation store, so it will look something like this https://github.com/anza-xyz/wincode/attestations/21526190

.github/workflows/publish.yml
buffalojoec: What's the harm in starting them at hashes now?

nbelenkov: None really, as long as we turn on depedanbot afterwards to actually bump those versions in case of security updates

nbelenkov: will add hashes

buffalojoec: Thanks for adding. Double-checked on my side!

PR #213, opened 2026-03-26
ci: adding dependabot settings

nbelenkov:
Add Dependabot configuration for Cargo and GitHub Actions dependencies.

Weekly version updates on Mondays, with Solana/Agave/SPL packages grouped into a single PR. Security updates fire immediately, regardless of the schedule

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
           - "solana-*"
           - "agave-*"
           - "spl-*"
-      all-dependencies:
+      third-party:
         patterns:
           - "*"
 


.github/dependabot.yml
```diff
@@ -0,0 +1,26 @@
...
+    schedule:
+      interval: weekly
+      day: monday
+    groups:
+      solana:
+        patterns:
+          - "solana-*"
+          - "agave-*"
+          - "spl-*"
+      all-dependencies:
```
buffalojoec: nit - how about:

```suggestion
      third-party:
```

PR #217, opened 2026-03-27
programs: Bump token-2022 to newest deployed version

joncinque:
#### Problem

While doing some testing on the token-wrap program with the newest crates, I ran into an `InvallidRealloc` error when calling into token-2022, which happens on the older version of the program and new runtime.

#### Summary of changes

Use the program currently deployed to mainnet in mollusk.


buffalojoec (approved):
```
$ solana-verify get-executable-hash programs/token/src/elf/token_2022.so
1ac923887d41793d89cc1f415dae6764aca5049c7eda6f6ae96ff01edeef4ed7
$ solana-verify get-program-hash -um TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb
1ac923887d41793d89cc1f415dae6764aca5049c7eda6f6ae96ff01edeef4ed7
```

PR #225, opened 2026-04-03
Add process_transaction_instructions to MolluskContext

rustopian:
### Problem

`MolluskContext` is missing wrappers for `process_transaction_instructions` and `process_and_validate_transaction_instructions`, making it impossible to get a `TransactionResult` (with `inner_instructions`) without manually assembling accounts, since `load_accounts_for_instructions` is private.

Our use case: We chain a `ComputeBudgetProgram::set_compute_unit_price` instruction before an instruction. The instruction emits events via CPI (inner instructions) that we need to inspect in tests. We cannot use `process_instruction_chain` on `MolluskContext`, since the swap's inner instructions get silently replaced by the compute budget instruction's empty inner instructions via `absorb()`, rather than extended.

### Fix
Add `process_transaction_instructions` and `process_and_validate_transaction_instructions` to `MolluskContext`, following the same pattern as the existing wrappers: load accounts via the store, delegate to `Mollusk`, then write back resulting accounts on success.

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -1839,6 +1839,15 @@ impl<AS: AccountStore> MolluskContext<AS> {
         }
     }
 
+    fn consume_transaction_result(&self, result: &TransactionResult) {
+        if result.program_result.is_ok() {
+            let mut store = self.account_store.borrow_mut();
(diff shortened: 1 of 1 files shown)


buffalojoec (approved):
Thanks!

PR #227, opened 2026-04-08
actions: add agave-4.0 branch to publish

buffalojoec:


nbelenkov (approved):
lgtm, for this to work, we would need to add `agave-4.0` to `prod` environment deployment branch rules

PR #228, opened 2026-04-08
Decouple `SysvarSerialize`

grod220:
Relates to this painful dep triangle: https://github.com/anza-xyz/agave/issues/10672 https://github.com/solana-program/stake/pull/343

A follow up after: https://github.com/anza-xyz/mollusk/pull/203#discussion_r3050218700

mollusk-svm failed [cargo publish --verify](https://github.com/anza-xyz/mollusk/actions/runs/24135033820) because solana-stake-interface's sysvar feature pulled in solana-sysvar v4.0 → solana-rent v4.0, conflicting with the v3 versions that Agave v4.0's runtime crates use.

`sysvar_account()`'s has a `SysvarSerialize` trait bound but only ever calls `bincode::serialize()`. If we remove that, we fix this dependency graph mess.

### DIFF
--- a/fuzz/fixture/Cargo.toml
+++ b/fuzz/fixture/Cargo.toml
@@ -26,7 +26,7 @@ solana-pubkey = { workspace = true }
 solana-rent = { workspace = true }
 solana-slot-hashes = { workspace = true }
 solana-stake-interface = { workspace = true }
-solana-sysvar.workspace = true
+solana-sysvar = { workspace = true }
 
 [build-dependencies]
 prost-build = { workspace = true }
--- a/harness/src/sysvar.rs
+++ b/harness/src/sysvar.rs
@@ -9,11 +9,11 @@ use {
     solana_program_runtime::sysvar_cache::SysvarCache,
     solana_pubkey::Pubkey,
     solana_rent::Rent,
+    solana_sdk_ids::sysvar::stake_history,
     solana_slot_hashes::{SlotHashes, MAX_ENTRIES as SLOT_HASHES_MAX_ENTRIES},
     solana_stake_interface::stake_history::{StakeHistory, StakeHistoryEntry},
     solana_sysvar::{
         self, last_restart_slot::LastRestartSlot, recent_blockhashes::RecentBlockhashes,
-        SysvarSerialize,
     },
     solana_sysvar_id::SysvarId,
 };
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -53,7 +53,7 @@ serde_json = "1.0.149"
 serde_yaml = "0.9.34"
 serial_test = "2.0"
 sha2 = "0.10.9"
-solana-account = "4.2"
+solana-account = "3.4"
```
grod220: These versions align with agave 4.0 deps now. Can be bumped back up when 4.1 libs come out.

harness/src/sysvar.rs
```diff
@@ -63,8 +63,7 @@ impl Default for Sysvars {
 }
 
 impl Sysvars {
-    fn sysvar_account<T: SysvarSerialize>(&self, sysvar: &T) -> (Pubkey, Account) {
-        let data = bincode::serialize::<T>(sysvar).unwrap();
+    fn sysvar_account(&self, pubkey: Pubkey, data: Vec<u8>) -> (Pubkey, Account) {
```
grod220, line 66: we could do

```rust
fn sysvar_account(&self, pubkey: Pubkey, sysvar: &impl serde::Serialize) -> (Pubkey, Account) {
    let data = bincode::serialize(sysvar).unwrap();
    ...
}
```
and make the call-sites simpler. But this would require making serde not an optional dep. That said, if bincode is being used, isn't it bringing it in regardless by default anyway?

fuzz/fixture/Cargo.toml
```diff
@@ -25,7 +25,7 @@ solana-keccak-hasher = { workspace = true }
 solana-pubkey = { workspace = true }
 solana-rent = { workspace = true }
 solana-slot-hashes = { workspace = true }
-solana-stake-interface = { workspace = true, features = ["bincode", "sysvar"] }
```
grod220: The critical feature to drop

grod220:
@buffalojoec review request 🙏

buffalojoec:
Sorry, closed by accident while I was reshuffling branches. Let's see if reopening "just works". 🙂

buffalojoec (commented):
This works fine, but can't we just avoid the dependency triangle if we stop using `solana-stake-interface` in Mollusk? It's only used in two places (both `sysvar` modules) for the `StakeHistory` sysvar information.

I'm curious why `StakeHistory` was moved into the Stake program's interface anyway, rather than it being a standalone sysvar crate like the others (ie. `solana-rent`/`solana-clock`). It would make this a lot easier.

grod220:
Yes, creating a new `solana-stake-history` crate would indeed simplify things. Though, it is an ecosystem-wide update that would take some time. But maybe the semver trick could come in handy there.

Think this PR can more immediately unblock things though. If you agree, can we move forward with this and I'll open another PR as a follow up with the new stake history sdk crate?

buffalojoec (approved):

PR #229, opened 2026-04-09
Add sbpf-debugger feature with GDB remote stub support

procdump:
**The problem**

There is no Mollusk integration with the SBPF debugger, meaning users couldn't interactively debug program execution from within Mollusk tests.

**What's changed**

- Add `sbpf-debugger` feature that starts a GDB remote stub on a configurable TCP port (`SBF_DEBUG_PORT`), allowing debugger clients to inspect registers, memory, set breakpoints, and step through SBPF execution
- Extract GDB RSP client helpers into a public `debugger` module for reuse by external tooling
- Refactor `elf_accounts_to_sha256` to deduplicate program IDs with HashSet and only persist ELF mappings when the debugger port is set
- Add CPI-aware debugger test

**Note**

Creating a new PR was easier than to reworking https://github.com/anza-xyz/mollusk/pull/200 which I can close once this PR is settled. In regards to the initially suggested SBF filtering, I think it's better to come in with a follow-up PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -541,6 +541,12 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
+
+[[package]]
+name = "bitflags"
 version = "2.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "843867be96c8daad0d758b57df9392b6d8d271134fce549de6ce169ff98a92af"
@@ -1528,6 +1534,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "gdbstub"
+version = "0.6.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4e02bf1b1a624d96925c608f1b268d82a76cbc587ce9e59f7c755e9ea11c75c"
+dependencies = [
+ "bitflags 1.3.2",
+ "cfg-if",
+ "log",
+ "managed",
+ "num-traits",
+ "paste",
+]
+
+[[package]]
 name = "generic-array"
 version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2016,6 +2036,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5e5032e24019045c762d3c0f28f5b6b8bbf38563a65908389bf7978758920897"
 
 [[package]]
+name = "managed"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ca88d725a0a943b096803bd34e73a4437208b6077654cc4ecb2947a5f91618d"
+
+[[package]]
 name = "memchr"
 version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2416,7 +2442,7 @@ version = "0.10.76"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "951c002c75e16ea2c65b8c7e4d3d51d5530d8dfa7d060b4776828c88cfb18ecf"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
  "cfg-if",
  "foreign-types",
  "libc",
@@ -2894,7 +2920,7 @@ version = "0.5.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ed2bf2547551a7053d6fdfafda3f938979645c44812fbfcda098faae3f1a362d"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
 ]
 
 [[package]]
@@ -2957,7 +2983,7 @@ version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6fe4565b9518b83ef4f91bb47ce29620ca828bd32cb7e408f0062e9930ba190"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
  "errno",
  "libc",
  "linux-raw-sys",
@@ -3627,7 +3653,7 @@ version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ddf67876c541aa1e21ee1acae35c95c6fbc61119814bfef70579317a5e26955"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
  "solana-account-info",
  "solana-instruction",
  "solana-instruction-error",
@@ -3811,7 +3837,7 @@ version = "4.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0ad62e1045c2347a0c0e219a6ceb0abfe904be622920996bfcac8d116fabe3c7"
 dependencies = [
- "bitflags",
+ "bitflags 2.11.0",
  "solana-pubkey 4.1.0",
 ]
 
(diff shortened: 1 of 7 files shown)


procdump:
@buffalojoec , will be happy to have your feedback.

Cargo.lock
buffalojoec, line 1: How'd we get all of these lockfile changes? I don't see any dependency changes. Is it all from `solana-program-runtime/sbpf-debugger`?

buffalojoec:
Sorry, closed by accident while I was reshuffling branches. Let's see if reopening "just works". 🙂

procdump:
Trying to fix the conflicts actually made it worse introducing changes unrelated to the debugger part.

harness/tests/register_tracing.rs
```diff
@@ -211,3 +222,200 @@ fn test_custom_register_tracing_callback() {
         );
     }
 }
+
+#[cfg(feature = "sbpf-debugger")]
+#[cfg(test)]
+mod debugger_tests {
```
buffalojoec: Does this need a submodule, or can it just be a single new test function? Seems like just moving the `use` statements out of the first test the top of the module would do it.

harness/src/lib.rs
```diff
@@ -561,20 +564,32 @@ pub trait InvocationInspectCallback {
         invoke_context: &InvokeContext,
         register_tracing_enabled: bool,
     );
+
+    fn as_any(&self) -> &dyn std::any::Any;
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
```
buffalojoec: This looks like it's only used in tests. Can we remove it from the trait?

harness/src/register_tracing.rs
```diff
@@ -22,6 +26,12 @@ impl Default for DefaultRegisterTracingCallback {
             // User can override default path with `SBF_TRACE_DIR` environment variable.
             sbf_trace_dir: std::env::var("SBF_TRACE_DIR").unwrap_or(DEFAULT_PATH.to_string()),
             sbf_trace_disassemble: std::env::var("SBF_TRACE_DISASSEMBLE").is_ok(),
+            // The port that will be used for debugging.
+            // Will invoke the debugger if set.
+            #[cfg(feature = "sbpf-debugger")]
+            sbf_debug_port: std::env::var("SBF_DEBUG_PORT")
+                .map(|port| port.parse::<u16>().ok())
+                .unwrap_or(DEFAULT_DEBUG_PORT),
```
buffalojoec: Can't you just `unwrap_or(None)` here or `unwrap_or_default`? I don't think the constant adds much value.

harness/src/register_tracing.rs
```diff
@@ -48,7 +58,32 @@ impl DefaultRegisterTracingCallback {
...
+        _instruction_accounts: &[InstructionAccount],
+        _invoke_context: &mut InvokeContext,
+    ) {
+        #[cfg(feature = "sbpf-debugger")]
+        {
+            if let Some(debug_port) = self.sbf_debug_port {
+                // Persist SHA-256 mapping for every ELF account.
+                // We need them later to judge what symbol object to
+                // load in the debugger client.
+                let _ = self.elf_accounts_to_sha256(
```
buffalojoec: The result is swallowed quietly here. Is that intentional?

harness/src/debugger.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/// Connects to the GDB stub with retries (100ms apart).
+/// Returns a buffered reader and a writer over the same TCP stream.
+pub fn stub_connect<A: ToSocketAddrs>(
+    stub_addr: A,
+    mut retries: usize,
+) -> Result<(BufReader<TcpStream>, TcpStream), std::io::Error> {
+    let (reader, writer) = loop {
+        retries -= 1;
```
buffalojoec: Should this be moved down? A value of `retries = 0` causes this to immediately wrap to `usize::MAX` and never break out.

Cargo.lock
procdump: Exactly and the `gdbstub` crate down the path introduces them.

harness/tests/register_tracing.rs
procdump: Removed the submodule and also I moved out the `use` statements splitting them per feature. The motivation - nesting in a single `use` statement would have resulted in too many `cfg` feature checks.

harness/src/lib.rs
procdump: Initially I thought it would be nice to have access to the inner type via downcasting but I believe this is better avoided and isn't worth it only for getting to `sbf_trace_dir` as was the case in the first place.

harness/src/register_tracing.rs
procdump: Should this happen I've added a simple eprint just to let the user be aware of the situation.

harness/src/debugger.rs
procdump: Good catch. Fixed.

harness/tests/register_tracing.rs
```diff
@@ -211,3 +225,180 @@ fn test_custom_register_tracing_callback() {
...
+            // let mut reply = stub_read_memory_chunked(
+            //     &mut writer,
+            //     &mut reader,
+            //     0x400000000,     // The input buffer of the program starts from here.
+            //     1 * 1024 * 1024, // Read 1MB just in case.
+            //     1024,            // Read in chunks of 1024 bytes.
+            // )?;
+            // let (deserialized_program_id, _, _) =
+            //     unsafe { solana_program_entrypoint::deserialize(reply.as_mut_ptr()) };
+            // assert_eq!(program_id, *deserialized_program_id);
```
buffalojoec: Did you mean to leave this commented code in?

harness/src/register_tracing.rs
```diff
@@ -48,7 +56,34 @@ impl DefaultRegisterTracingCallback {
...
     }
 
-    pub fn handler(
+    pub fn pre_handler(
+        &self,
+        _mollusk: &Mollusk,
+        _program_id: &Pubkey,
+        _instruction_data: &[u8],
+        _instruction_accounts: &[InstructionAccount],
+        _invoke_context: &mut InvokeContext,
```
buffalojoec: Any reason we need to keep these prefixed with underscores?

harness/src/debugger.rs
```diff
@@ -0,0 +1,193 @@
...
+    // Read till the # character.
+    reader.read_until(b'#', &mut buf)?;
+    // Then read exactly 2 bytes representing the checksum.
+    let mut cbuf = [0];
+    let _ = reader.read(&mut cbuf)?;
+    let _ = buf.write(&cbuf)?;
+    let _ = reader.read(&mut cbuf)?;
+    let _ = buf.write(&cbuf)?;
+    let reply = String::from_utf8_lossy(&buf).to_string();
+    // eprintln!("gdbstub reply: {}", reply);
```
buffalojoec: Maybe we can drop this commented line as well?

README.md
```diff
@@ -640,5 +645,6 @@ mollusk.invocation_inspect_callback =
     Box::new(register_tracing::DefaultRegisterTracingCallback {
         sbf_trace_dir: std::env::var("SBF_TRACE_DIR").unwrap(),
         sbf_trace_disassemble: std::env::var("SBF_TRACE_DISASSEMBLE").is_ok(),
+        sbf_debug_port: std::env::var("SBF_DEBUG_PORT").unwrap(),
```
buffalojoec: Isn't this actually an `Option<u16>`? I think this snippet has to match what you put in the `register_tracing` module in this PR for `sbf_debug_port`.

buffalojoec (commented):
Just a few small cleanup items, but this is ready to rip. Nice work! Really cool to have this feature!

harness/tests/register_tracing.rs
procdump: I kept it just for illustrating the ability users would have right across the fence in terms of inspection - I believe it's better removed now.

README.md
procdump: Right - fixed.

harness/src/register_tracing.rs
procdump: I reworked it - [953fa16](https://github.com/anza-xyz/mollusk/pull/229/commits/953fa16e3112301ab16bec948917cdf948150fb2)

buffalojoec (approved):

buffalojoec:
I'll see if I can get this published in `0.12.1-agave-4.0`! Maybe later today.

PR #239, opened 2026-04-15
harness: convert to use SVMFeatureSet

buffalojoec:
Mollusk is an SVM-level testing harness. There's no reason anyone should have to mess with Agave features that toggle nothing in Mollusk's test environment.

However, working with fixtures and the Mollusk CLI is actually a lot easier when you can activate/deactivate features by 32-byte address (I believe). So, we'll leave `FeatureSet` alone for the `fuzz` APIs.


febo (approved):
Looks good!

PR #249, opened 2026-05-12
ci: Move SLSA attestation before publish

nbelenkov:
The last publish run failed because `actions/attest-build-provenance` couldn't find any `.crate` files in `target/package/`. Turns out cargo 1.91 changed behavior: it no longer keeps `.crate` tarballs around as final artifacts after `cargo publish`. We bumped the toolchain from 1.88 → 1.93 along with the agave 4.0 deps bump, which is when this broke.

Fix: package first, attest the `.crate` files, then publish.

- New `make package` target uses cargo's multi-package mode to generate all 10 `.crate` files up front
- Workflow reordered: package → attest → auth → publish
- Pulled the crate list into a shared `PUBLISH_CRATES` make var so `package` and `publish` aren't duplicating it


nbelenkov:
@buffalojoec this should fix the failing action

.github/workflows/publish.yml
```diff
@@ -77,6 +77,15 @@ jobs:
         with:
           cache: true
           version: ${{ steps.solana.outputs.version }}
+      - name: Package
+        shell: bash
+        env:
+          ARGS: "--locked"
+        run: make package
```
buffalojoec, line 84: Should we also make this job and `Publish` use SHA-pinned containers instead of falling back to default? This way, we have max determinism in the filesystem.

buffalojoec (commented):
The only thing that kind of stinks here is that we're relying on the generated `.crates` dir during publish to be 100% byte-for-byte with the one from `package`, otherwise the attestation step is worthless.

I suppose by using `--locked` for both jobs, we should be fine, but I still find it so strange that there isn't better SLSA support for cargo native tooling.

Also, you can rebase to fix the audit CI failure after https://github.com/anza-xyz/mollusk/pull/250

.github/workflows/publish.yml
nbelenkov: Do you mean the whole action to run on a sha pinned container like this? 
```
runs-on: ubuntu-latest
container:
  image: ubuntu@sha256:<some-digest> 
 ```

nbelenkov:
Yeah, the way crates have it set up is not ideal. We are very much trusting that the crates are bit-by-bit identical, there have been attempts to get this stuff more properly integrated into cargo https://github.com/rust lang/cargo/issues/12661, but with not much success. 

In practice, it somewhat holds because cargo 1.91+ normalizes tarball timestamps and file ordering ([rust-lang/cargo#16242](https://github.com/rust-lang/cargo/pull/16242)) and `--locked` pins dep resolution, but you're right that we're trusting it rather than proving it.

If we want to actually prove it, the cleanest check is: sha256 the `.crate` files after `make package`, then after publish, re-download each crate from `crates.io` and diff the digests. Which is very painful, ngl. It's a side effect that SLSA was mainly made for binaries: you would compare the sha256s at attestation and at release, but cargo publish no longer preserves the .crate file, which is annoying.

.github/workflows/publish.yml
buffalojoec: Yeah, but maybe that's not all that useful after all. What do you think?

buffalojoec (dismissed):

buffalojoec (approved):

PR #255, opened 2026-05-21
Agave 4.0

buffalojoec:
#### Problem
We've been maintaining an `agave-4.0` branch while the Agave 4.0 crates were still in prerelease status, but now they're fully published.

#### Summary of Changes
Merge `agave-4.0` into `main`, upgrading Mollusk to use the brand new crates.

Version bumps aren't automatic on this repo, so I've included one commit at the end to bump the version. I'll tag the merge commit as the next major.


buffalojoec:
@grod220 I wanted to request your review but you need to be added to the Mollusk group on GitHub (via devops).

joncinque (approved):
Looks good to me!

.github/workflows/publish.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   pre_publish:
     name: Prepublish
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/main'
+    if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/agave-4.0'
```
febo, line 12: @buffalojoec Just having a quick look here before approving the publishing. Wondering whether you intended to merge this to main as well. Since the `agave-4.0` branch probably won't be used anymore, maybe it is not needed?

buffalojoec: Ah yeah, it can come out. I'll do that in follow-up. Thanks!

PR #257, opened 2026-05-22
Default rent sysvar to SIMD-0194 layout

grod220:
[SIMD-0194](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0194-deprecate-rent-exemption-threshold.md) is activated on mainnet. The rent sysvar's first 8 bytes now encode `lamports_per_byte` directly (6960), not the legacy `lamports_per_byte_year` (3480).

Mollusk still writes the legacy layout. Readers that adopt the new layout (pinocchio-system 0.6.1's `CreateAccountAllowPrefund::with_minimum_balance()`) read 3480 as the per-byte cost. This underfunds created accounts by half and causes errors:

```
Lamport balance below rent-exempt threshold
```


grod220:
@buffalojoec can you review?  

Would be great to get a patch release after this 🙏

buffalojoec (commented):
I think this only works because cargo has autoresolved `solana-rent` to 3.1.0. Can we bump it to 3.1.0 in the workspace manifest?

buffalojoec (approved):

grod220:
Thank you! Can you publish a patch? 🙏

buffalojoec:
> Thank you! Can you publish a patch? 🙏

https://github.com/anza-xyz/mollusk/releases/tag/0.13.1

PR #260, opened 2026-06-13
feat: token22 extensions

sonicfromnewyoke:
### Problem

token-2022 helpers only build base-layout mints and token accounts - copy-pasted from token helpers (except for the program ID and loader)
so, there is no way to set up Token-2022 extension state (the entire point of Token-2022) without hand-rolling TLV bytes
taking into account that meteora-dbc released v0.2 last week which hevily relies on transfer-hooks - better suport of them will be nice to have

### Summary of Changes

- add mint and token account extension enums
- add extension-aware builders for them
- extend tests

#### Example: mint with a transfer hook

```rust
use {
    mollusk_svm::Mollusk,
    mollusk_svm_programs_token_2022::{
        self as token2022, AccountState, MintExtension, TokenAccount, TransferHook,
    },
    solana_program_option::COption,
    solana_pubkey::Pubkey,
};

let mut mollusk = Mollusk::default();
token2022::add_program(&mut mollusk);

let authority = Pubkey::new_unique();
let hook_program = Pubkey::new_unique();
let mint = Pubkey::new_unique();
let owner = Pubkey::new_unique();

// Mint with the transfer hook extension
let mint_account = token2022::create_account_for_mint_with_extensions(
    token2022::Mint {
        mint_authority: COption::Some(authority),
        supply: 0,
        decimals: 9,
        is_initialized: true,
        freeze_authority: COption::None,
    },
    &[MintExtension::TransferHook(TransferHook {
        authority: Some(authority).try_into().unwrap(),
        program_id: Some(hook_program).try_into().unwrap(),
    })],
);

// Token account with the required `TransferHookAccount` extension,
// derived automatically from the mint
let token_account = token2022::create_account_for_token_account_with_extensions(
    TokenAccount {
        mint,
        owner,
        amount: 0,
        delegate: COption::None,
        state: AccountState::Initialized,
        is_native: COption::None,
        delegated_amount: 0,
        close_authority: COption::None,
    },
    &token2022::required_account_extensions_for_mint(&mint_account.data),
);
```

cc @buffalojoec


programs/token/src/token2022.rs
```diff
@@ -1,16 +1,202 @@
...
+            }
+            Self::TransferHook(extension) => {
+                *state.init_extension::<TransferHook>(true).unwrap() = *extension;
+            }
+            Self::MetadataPointer(extension) => {
+                *state.init_extension::<MetadataPointer>(true).unwrap() = *extension;
+            }
+            Self::TokenMetadata(extension) => {
+                state.init_variable_len_extension(extension, true).unwrap();
+            }
```
buffalojoec: You partition in the account-builder helper (`create_account_for_mint_with_extensions`) between fixed- and variable-length extensions. If someone calls this and they need to resize, it won't work, right? Maybe we just stub this out with `unreachable!` or just do a `panic!`.

buffalojoec (changes requested):
Looks good! I think if we're going to do this, let's split it up and have a dedicated Token-2022 version, so we don't burden anyone with Token-2022-related dependencies if they only need `Mint`/`Account`.

* `mollusk-svm-programs-token` - just the `Mint` and `Account` as was pre-existing
* `mollusk-svm-programs-token-2022` imports from above ^ and maybe re-exports stuff to avoid breakage

programs/token/src/token2022.rs
sonicfromnewyoke: good catch. Done

sonicfromnewyoke:
> Looks good! I think if we're going to do this, let's split it up and have a dedicated Token-2022 version, so we don't burden anyone with Token-2022-related dependencies if they only need `Mint`/`Account`.
> 
>     * `mollusk-svm-programs-token` - just the `Mint` and `Account` as was pre-existing
> 
>     * `mollusk-svm-programs-token-2022` imports from above ^ and maybe re-exports stuff to avoid breakage

such kind of design fills a little bit awkward, because stands out of how it is done for a regular token and ATA programs. However - done 🫡

buffalojoec:
> such kind of design fills a little bit awkward, because stands out of how it is done for a regular token and ATA programs. However - done 🫡

Yeah I know, but I plan to move ATA in the future as well. Can you sweep the code from `programs/token/src/token_2022.rs` into the new crate and re-export from the original?

sonicfromnewyoke:
> > such kind of design fills a little bit awkward, because stands out of how it is done for a regular token and ATA programs. However - done 🫡
> 
> Yeah I know, but I plan to move ATA in the future as well. Can you sweep the code from `programs/token/src/token_2022.rs` into the new crate and re-export from the original?

done, re-exporting token-2022 functionality without extensions

buffalojoec (approved):
Looks great! Thanks!

PR #266, opened 2026-06-23
ci: harden the publish pipeline with ownership + dry-run gates

buffalojoec:
Add two pre-flight gates to the publish workflow so a release can't fail partway and poison crates.io (as happened for token-2022 in 0.13.2):

- verify_crate_owners: a tokenless check (scripts/verify-crate-owners.sh) that every publishable crate already exists on crates.io and is owned by the trusted anza-team. Runs in parallel with prepublish.
- publish_dry_run: a full `cargo publish --workspace --dry-run` that builds and packages every crate before any real upload.

Both gate the real publish job.

Also modernize the publish recipes and de-duplicate CI setup:

- publish/package use `cargo {publish,package} --workspace`, dropping the hand-maintained PUBLISH_CRATES list and the per-crate loop + sleeps; cargo orders the crates and waits for index propagation itself.
- the real publish reads the token from CARGO_REGISTRY_TOKEN.
- a publish-setup composite action holds the shared cache/protoc/solana setup reused across the publish jobs.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,71 +12,58 @@ jobs:
     if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/agave-4.0'
     permissions:
       contents: read
-
     steps:
       - uses: actions/checkout@v7
-      - uses: actions/cache@v5
+      - uses: ./.github/actions/publish-setup
         with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-          key: cargo-build-sbf-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            cargo-build-sbf-
-      - name: Install protoc
-        run: sudo apt-get install -y protobuf-compiler
-      - name: Get nightly toolchain version
-        id: nightly
-        run: echo "version=$(make nightly-version)" >> $GITHUB_OUTPUT
-      - uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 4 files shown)


scripts/verify-crate-owners.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+#
+# Verify that every publishable crate in the workspace already exists on
+# crates.io and is owned by the trusted Anza team.
+#
+# Adapted from ci/check-crates.sh in anza-xyz/agave.
+
+set -euo pipefail
```
joncinque, line 8: nit: up to you, but this script currently runs on whatever directory the shell is in, and not necessarily the workspace root. 

It might be useful to `cd` to the workspace root to make sure it works as intended

.github/workflows/publish.yml
```diff
@@ -12,71 +12,58 @@ jobs:
...
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+      - uses: ./.github/actions/publish-setup
+      - name: Dry-run publish
+        shell: bash
+        env:
+          ARGS: "--locked"
+        run: make publish DRY_RUN=--dry-run
```
joncinque, line 51: I never knew this syntax was possible, very neat!

joncinque (dismissed):
Looks good to me! Just one nit that you can take or leave

PR #271, opened 2026-07-06
Migrate stake-interface to stake-history crate

grod220:
Follows work from https://github.com/anza-xyz/solana-sdk/pull/766


Cargo.toml
```diff
@@ -89,7 +89,7 @@ solana-sdk-ids = "3.1.0"
 solana-secp256k1-program = "3.0.1"
 solana-secp256r1-program = "3.0.0"
 solana-slot-hashes = "3.0.1"
-solana-stake-interface = "2.0.2"
+solana-stake-history = { version = "1.0.0", features = ["serde"] }
```
buffalojoec: nit: can you just declare the version here and then enable the features in the crates that use them (ie. `harness`)?

harness/src/sysvar.rs
```diff
@@ -420,8 +420,8 @@ mod tests {
             &sysvars.slot_hashes
         );
         assert_eq!(
-            sysvar_cache.get_stake_history().unwrap().deref(),
-            &sysvars.stake_history
+            bincode::serialize(sysvar_cache.get_stake_history().unwrap().deref()).unwrap(),
+            bincode::serialize(&sysvars.stake_history).unwrap(),
```
buffalojoec, line 426: nit: maybe just include a TODO here saying that once Agave's `SysvarCache` is also using the new lib, this can go back to direct comparison.

buffalojoec (commented):
Looks great! Two tiny things and this can land.

buffalojoec (dismissed):
Looks great! Might need a rebase after #272 but good to go from my side.

buffalojoec (approved):

PR #276, opened 2026-07-10
deps: bump to agave 4.2.0-beta.0

buffalojoec:

PR #277, opened 2026-07-10
deps: Agave with ABIv2 support

febo:
### Problem

Agave will support ABIv2 runtime in a future version and currently it is not possible to test ABIv2 programs in mollusk.

### Solution

Update the agave dependencies to use the current work-in-progress ABIv2 Agave [fork](https://github.com/anza-xyz/agave-runtime). This allows testing ABIv2 programs – see https://github.com/febo/abiv2.

### DIFF
--- a/harness/src/program.rs
+++ b/harness/src/program.rs
@@ -88,7 +88,7 @@ impl ProgramCache {
         enable_register_tracing: bool,
     ) -> Self {
         let me = Self {
-            cache: Rc::new(RefCell::new(ProgramCacheForTxBatch::default())),
+            cache: Rc::new(RefCell::new(ProgramCacheForTxBatch::new(1))),
             entries_cache: Rc::new(RefCell::new(HashMap::new())),
             program_runtime_environment: create_program_runtime_environment(
                 feature_set,
@@ -175,9 +175,8 @@ impl ProgramCache {
                     loader_key,
                     environment,
                     0,
-                    0,
+                    1,
                     elf,
-                    elf.len(),
                     &mut LoadProgramMetrics::default(),
                 )
                 .unwrap(),
@@ -247,7 +246,6 @@ impl Builtin {
     fn program_cache_entry(&self) -> Arc<ProgramCacheEntry> {
         Arc::new(ProgramCacheEntry::new_builtin(
             0,
-            self.name.len(),
             self.register_fn,
         ))
     }
--- a/harness/tests/elf_loading.rs
+++ b/harness/tests/elf_loading.rs
@@ -1,9 +1,6 @@
 //! Tests for loading SBPF ELFs.
 
-use {
-    mollusk_svm::{program::ProgramCache, Mollusk},
-    solana_pubkey::Pubkey,
-};
+use {mollusk_svm::Mollusk, solana_pubkey::Pubkey};
 
 #[test]
 fn test_sbpf_v0_elf_loads() {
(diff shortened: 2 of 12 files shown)


buffalojoec (commented):
Looks good, but some of this churn will come out if you rebase onto #276.
I created a branch `abiv2` if you want to target that one instead. It's based on the `agave-4.2.0-beta` pre-release branch.
https://github.com/anza-xyz/mollusk/tree/abiv2

febo:
> Looks good, but some of this churn will come out if you rebase onto #276. I created a branch `abiv2` if you want to target that one instead. It's based on the `agave-4.2.0-beta` pre-release branch. https://github.com/anza-xyz/mollusk/tree/abiv2

<del>It should be already based on #276 since I started a "fresh" one from main after that PR was merged.</del> I changed the PR tp target `abiv2` branch.

update: I see what you mean, #276 was not targeting main.

buffalojoec:
I think you just need a rebase onto `abiv2` @febo then we can cook.

febo:
> I think you just need a rebase onto `abiv2` @febo then we can cook.

Done! Maybe we need to allow CI to run on this branch.

harness/tests/elf_loading.rs
```diff
@@ -33,9 +30,9 @@ fn test_sbpf_v2_elf_loads() {
 }
 
 #[test]
-fn test_sbpf_v3_feature_is_active_by_default() {
+fn test_program_runtime_abiv2_feature_is_active_by_default() {
     let mollusk = Mollusk::default();
-    assert!(mollusk.feature_set.enable_sbpf_v3_deployment_and_execution);
+    assert!(mollusk.feature_set.program_runtime_abiv2);
 }
```
buffalojoec: nit: this is kind of an awkward place for this test, since the test module is focused on ELF loading. I won't block the PR, though.

harness/src/lib.rs
```diff
@@ -1030,62 +1030,41 @@ impl Mollusk {
...
+        let raw_result = invoke_context
+            .process_message(sanitized_message, &mut timings, &mut compute_units_consumed)
+            .map_err(|(index, err)| TransactionError::InstructionError(index, err));
+
+        #[cfg(feature = "invocation-inspect-callback")]
+        self.invocation_inspect_callback.after_invocation(
+            self,
+            &invoke_context,
+            self.enable_register_tracing,
+        );
```
buffalojoec: I don't think the callback is going to work the same when executing a full transaction. Previously, the callback was tapped right before and immediately after each instruction. We might have to keep the old code here somewhere that iterates over the message instructions manually if `invocation_inspect-callback` is enabled.

The only catch is I believe `process_precompile` is now private, so we'd need to make it public under DCOU on the runtime fork.

harness/tests/elf_loading.rs
febo: Is there a better place for it? Maybe we create an abiv2 specific module?

harness/src/lib.rs
febo: Yeah, this one was weird to fix the compilation issues since `process_precompile` became private. To make the change to be under DCOU, we should do it to agave main first, right?

harness/tests/elf_loading.rs
buffalojoec: Yep, that works1

harness/src/lib.rs
buffalojoec: Yeah, just link here in the PR, since we'll need to provide context. This change will happen on Mollusk's main release train, too, so it's not just an ABIv2 thing.

harness/src/program.rs
```diff
@@ -88,7 +88,7 @@ impl ProgramCache {
         enable_register_tracing: bool,
     ) -> Self {
         let me = Self {
-            cache: Rc::new(RefCell::new(ProgramCacheForTxBatch::default())),
+            cache: Rc::new(RefCell::new(ProgramCacheForTxBatch::new(1))),
```
febo, line 91: This was needed because of https://github.com/anza-xyz/agave/pull/14168

harness/src/program.rs
```diff
@@ -175,9 +175,8 @@ impl ProgramCache {
                     loader_key,
                     environment,
                     0,
-                    0,
+                    1,
```
febo, line 178: Same as above.

buffalojoec (approved):

PR #286, opened 2026-07-28
harness!: Add payer account to `process_transaction_instructions`

### LINKED ISSUE #283: Add helpers that accept a specific payer account
Currently, there is no direct way to specify the payer account – it is set by default to `None` [here](https://github.com/anza-xyz/mollusk/blob/1a214a04cb892746ff80f4bb8a4261b488c00dc8/harness/src/compile_accounts.rs#L18). In ABIv2, it is now possible to access the payer account so it would be interesting to provide helpers that take a payer account.

febo:
### Problem

As noted in https://github.com/febo/abiv2/pull/2, the transaction fee payer in ABIv2 is accessible to programs without being included in instructions directly. This allows programs to optimize signer validation by not requiring an `AccountMeta` for the signer, since the fee payer is always a signer. Currently this is not possible bacause there is no way to specify the fee payer account.

### Solution

Update `process_transaction_instructions*` helpers to accept a payer address. Decided to add to only the `process_transaction*` given that the payer represents the transaction fee payer.

Fixes #283

cc: @deanmlittle

### DIFF
--- a/harness/src/lib.rs
+++ b/harness/src/lib.rs
@@ -1323,19 +1323,8 @@ impl Mollusk {
 
     /// Process multiple instructions using a single shared transaction context.
     ///
-    /// This is the same as
-    /// [`Self::process_transaction_instructions_with_payer`], but without a
-    /// payer. In this case, the payer is assumed to be the first account in the
-    /// `accounts` slice.
-    pub fn process_transaction_instructions(
-        &self,
-        instructions: &[Instruction],
-        accounts: &[(Pubkey, Account)],
-    ) -> TransactionResult {
-        self.process_transaction_instructions_with_payer(instructions, accounts, None)
-    }
-
-    /// Process multiple instructions using a single shared transaction context.
+    /// When a transaction fee `payer` is not provided, it is assumed to be the
+    /// first account in the `accounts` slice.
     ///
     /// This API is the closest Mollusk offers to a transaction. All
     /// instructions are processed in the same message using the same
@@ -1351,7 +1340,7 @@ impl Mollusk {
     /// * `program_result`: The result code of the last program's execution and
     ///   its index.
     /// * `resulting_accounts`: The resulting accounts after all instructions.
-    pub fn process_transaction_instructions_with_payer(
+    pub fn process_transaction_instructions(
         &self,
         instructions: &[Instruction],
         accounts: &[(Pubkey, Account)],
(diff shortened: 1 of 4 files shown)


buffalojoec (commented):
I could go either way, but why introduce a new API method instead of just breaking the lib and adding the `payer` arg to `process_transaction_instructions`?

febo:
> I could go either way, but why introduce a new API method instead of just breaking the lib and adding the `payer` arg to `process_transaction_instructions`?

Happy either way – I was trying to avoid a breaking change, but if that is not a problem, agree that it would be better to just change the existing method to expect an extra parameter.

deanmlittle:
> I could go either way, but why introduce a new API method instead of just breaking the lib and adding the `payer` arg to `process_transaction_instructions`?

if we do this, i'd go for an `Option<Address>` as most tests don't require it.

febo:
> > I could go either way, but why introduce a new API method instead of just breaking the lib and adding the `payer` arg to `process_transaction_instructions`?
> 
> if we do this, i'd go for an `Option<Address>` as most tests don't require it.

Yeah, currently it is an `Option`.

harness/src/lib.rs
```diff
@@ -1522,7 +1528,7 @@ impl Mollusk {
         accounts: &[(Pubkey, Account)],
         checks: &[Check],
     ) -> TransactionResult {
```
buffalojoec, line 1534: Can you also add the `payer` arg here? Since it's a complementary API to `process_transaction_instructions`.

harness/src/lib.rs
```diff
@@ -1940,7 +1946,7 @@ impl<AS: AccountStore> MolluskContext<AS> {
         let accounts = self.load_accounts_for_instructions(instructions.iter());
         let result = self
             .mollusk
-            .process_transaction_instructions(instructions, &accounts);
+            .process_transaction_instructions(instructions, &accounts, None);
```
buffalojoec: I think the API for the MolluskStore shouldn't move out-of-step with the Mollusk API. Let's add `payer` to this API, too.

buffalojoec (approved):

harness/src/lib.rs
febo: Missed that one – added in https://github.com/anza-xyz/mollusk/pull/286/commits/ea06a26f5abf8cd5cbc762a8d05548b7eb031db3

PR #288, opened 2026-07-31
ci: Fix version in abiv2 branch

febo:
### Problem

PR #277 wrongly updated the solana cli version to `4.3.0-alpha.0`, which does not exist. It also update the workspace version to `0.14.0-abiv2` but did not change the individual crate versions in the dependency list.

### Solution

Update the version to the latest `4.2.0-rc.0` and fix the version of each individual crate.

Note: the "Rustfmt" error on CI is expected – it is fixed in #289

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SHELL := /usr/bin/env bash
 NIGHTLY_TOOLCHAIN := nightly-2026-04-11
-SOLANA_VERSION := 4.3.0-alpha.0
+SOLANA_VERSION := 4.0.0
 
 
 .PHONY: \
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1913,7 +1913,7 @@ dependencies = [
 
 [[package]]
 name = "mollusk-svm"
-version = "0.14.0-agave-4.2.0-beta.0"
+version = "0.14.0-abiv2"
 dependencies = [
  "agave-feature-set",
  "agave-precompiles",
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 SHELL := /usr/bin/env bash
 NIGHTLY_TOOLCHAIN := nightly-2026-04-11
-SOLANA_VERSION := 4.3.0-alpha.0
+SOLANA_VERSION := 4.0.0
```
buffalojoec, line 3: 4.3.0-alpha may not exist, but the runtime fork is based on 4.3, which means we at least want to use the 4.2 CLI, right?

febo: Yeah, I was not sure. If I bump to 4.2, then there are some CUs that need updating on the tests. Maybe we should set the `--tools-version` instead of the CLI version. Or use `cargo-build-sbf` directly. This is used just to compile the programs right?

buffalojoec: Yeah that's correct

febo: I would suggest keeping with `4.0.0` for now and then switch to use [`cargo-build-sbf`](https://crates.io/crates/cargo-build-sbf) in a separate PR. What do you think?

buffalojoec: Yeah, works for me. I bumped it on the `4.2-beta` branch, but that's fine. This is a development branch here.

buffalojoec (approved):
Acking your note on rustfmt and merging on red.

PR #298, opened 2026-08-12
fix: derive Firedancer fixture CPI stack depth from its feature set

latent-9:
## What

`parse_fixture_context` (the Firedancer fixture replay path) rebuilds the `ComputeBudget` with `ComputeBudget::new_with_defaults(true)`, hardcoding the SIMD-0268 (`raise_cpi_nesting_limit_to_8`) flag to active regardless of the fixture's own feature set.

## Why

`new_with_defaults(simd_0268_active)` sets `max_instruction_stack_depth` to 9 (8 nested CPIs) when the flag is on and 5 (4 nested CPIs) when it is off. The fixture's real feature set is available in the same function (`epoch_context.feature_set`) and is used a few lines later for `ParsedFixtureContext.feature_set`, but the compute budget ignored it.

So for any Firedancer fixture where SIMD-0268 is inactive, Mollusk replays with a stack depth of 9 while the agave/Firedancer runtime uses 5. A program performing 5 to 8 nested CPIs fails on-chain with `MaxInstructionStackDepth` but succeeds under Mollusk fixture replay, so conformance silently passes for transactions the real runtime rejects.

The sibling non-Firedancer path already derives the flag from the feature set (`fuzz/fixture/src/context.rs`, `feature_set.is_active(&raise_cpi_nesting_limit_to_8::id())`), with tests asserting `new_with_defaults(false)` gives a depth of 5.

## Change

Derive `raise_cpi_nesting_limit_to_8` from the fixture's feature set (already parsed in the same function) and pass it to `new_with_defaults`. The unrelated `Mollusk::new_inner` default keeps `true` because it enables all features.

## Test

`test_parse_fixture_context_honors_fixture_feature_set` builds a fixture whose feature set does not activate SIMD-0268 and asserts the parsed compute budget matches `ComputeBudget::new_with_defaults(false)` (depth 5). It fails before the change (hardcoded depth 9) and passes after; the existing Firedancer tests still pass.

### DIFF
--- a/harness/src/fuzz/firedancer.rs
+++ b/harness/src/fuzz/firedancer.rs
@@ -140,9 +140,10 @@ pub(crate) fn parse_fixture_context(context: &FuzzContext) -> ParsedFixtureConte
         epoch_context,
     } = context;
 
+    let feature_set = epoch_context.feature_set.runtime_features();
     let compute_budget = ComputeBudget {
         compute_unit_limit: *compute_units_available,
-        ..ComputeBudget::new_with_defaults(true)
+        ..ComputeBudget::new_with_defaults(feature_set.raise_cpi_nesting_limit_to_8)
     };
 
     let accounts = accounts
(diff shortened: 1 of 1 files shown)


buffalojoec (approved):
Thanks!

PR #299, opened 2026-08-13
misc housekeeping

buffalojoec:
Just moving some code around to clean up `lib.rs`.

@febo requesting your review to make sure this won't make rebasing your misc branches too annoying.


febo (approved):
Looks good!

PR #302, opened 2026-08-18
harness: make loader v1 a default builtin

buffalojoec:
There's really no reason this one can't be a default builtin. It shares the same function as the other BPF Loader programs.

I'm also _pretty sure_ this is non-breaking?


grod220 (approved):
🚢

PR #306, opened 2026-08-29
fix: include return data in fixture hashes

kimjune01:
## What

Include `InstrEffects::return_data` when computing deterministic Mollusk fixture hashes.

## Why

Return data was added to fixture effects after deterministic hashing was introduced, but the hash function was not updated.

This meant two otherwise identical fixtures produced the same hash when only their return data differed. Because fixture hashes determine file naming, distinct fixtures could be deduplicated or overwrite one another.

The Firedancer fixture format already includes return data in its equivalent hash.

## Change

Hash `return_data` alongside the other `InstrEffects` fields.

Add a regression test that changes only return data and verifies that the fixture hash also changes.

## Test

- `cargo test -p mollusk-svm-fuzz-fixture`
- `cargo clippy -p mollusk-svm-fuzz-fixture --all-targets -- -D warnings`
- `cargo fmt --all -- --check`


buffalojoec (approved):
Thanks! This must have gotten missed.

PR #308, opened 2026-09-04
fix: correctly set sysvar ixns account is_writable / is_signer flags 

soundsonacid:
the ixns sysvar in agave uses transaction-level `SVMMessage` trait impls https://github.com/anza-xyz/agave/blob/master/svm/src/account_loader.rs#L653

these return highest txn-level perms (as opposed to per ixn), so if acc is writable in ixn 1, readonly in ixn 2, the `BorrowedInstruction` for 2 will still show acc as writable in 2 (or like, the "effective privilege" for acc instead of the actual intent)

mollusk reads the actual flag on the acc on each ixn https://github.com/anza-xyz/mollusk/blob/main/harness/src/instructions_sysvar.rs#L8

for the transaction-shaped harness processors this will return divergent behavior from same set of ixns put into a txn on mainnet

also make ixns sysvar creation in build_transaction_accounts not dead code by removing creation in get_fallback_accounts.

cc @cavemanloverboy for initial finding

### DIFF
--- a/harness/src/compile_accounts.rs
+++ b/harness/src/compile_accounts.rs
@@ -77,8 +77,7 @@ fn build_transaction_accounts(
                     return (*key, AccountSharedData::from(provided_account.clone()));
                 }
                 // We will never have a fallback for the Instructions sysvar.
-                let (_, account) =
-                    crate::instructions_sysvar::keyed_account(message);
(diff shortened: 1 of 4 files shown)


buffalojoec (approved):
Nice, good fix!

PR #317, opened 2026-09-22
harness: check rent exemption by default

buffalojoec:
Currently, checking for rent exemption is opt-in, and this has been a footgun for developers.

With the next release, we'll make it the default behavior, for all harnesses, and offer an opt-out config:

```rust
mollusk.config.rent_exempt_checks = false;
```


febo:
This is a great idea!

I tried to test another scenario where there is an account with data below the minimum exempt:
```rust
#[test]
#[should_panic(expected = "left in a rent state the runtime rejects")]
fn test_funding_new_account_with_data_below_minimum_fails() {
    let sender = Pubkey::new_unique();
    let recipient = Pubkey::new_unique();

    let instruction = transfer(&sender, &recipient, 1);

    let mut recipient_account = Account::default();
    recipient_account.lamports = exempt_minimum() - 1;
    // Add some data to the account.    
    recipient_account.data = vec![4; 100];

    let accounts = [
        (sender, system_account(exempt_minimum() * 2)),
        (recipient, recipient_account),
    ];

    Mollusk::default().process_and_validate_instruction(&instruction, &accounts, &[]);
}
```
But this does not panic as expected.

buffalojoec:
Is this not SIMD-0392 rules, which assume no pre-account can be rent-paying?

febo:
> Is this not SIMD-0392 rules, which assume no pre-account can be rent-paying?

Yes, you are right. I was trying to add a test where an account is left with data but not enough lamports. Probably needs a custom program to do that.

harness/tests/rent_checks.rs
```diff
@@ -0,0 +1,165 @@
...
+        (bob, system_account(exempt_minimum() * 2)),
+    ];
+
+    let checks: &[Check] = &[];
+
+    Mollusk::default().process_and_validate_instruction_chain(
+        &[(&instruction_one, checks), (&instruction_two, checks)],
+        &accounts,
+    );
+}
```
febo, line 166: Could we add a test with an account with data and rent exempt that has its lamports zeroed by an instruction? E.g.:
```
Pre-instruction:

Account A:
  - data len: 100
  - lamports: 0.00115824

Post-instruction:

Account A:
  - data len: 100
  - lamports: 0
```
This should not trigger the rent exempt check.

febo (commented):
Looks good! Just added a suggestion for one more test.

harness/tests/rent_checks.rs
buffalojoec: Done in https://github.com/anza-xyz/mollusk/pull/317/changes/74bab07c3a62aebb45f166976eb887c7b27c90c4!

